PR #1, opened 2016-01-29
Add profile.proto and third party directory

matthewloring:
Proto imported from https://github.com/google/pprof

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 build
 node_modules
-third_party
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
Since the file doesn't contain a copyright, can you include the LICENSE file from the upstream project into the `third_party/proto` directory. Also include a README.google file documenting the upstream location this file was copied from.

/cc @rauls5382

matthewloring:
@ofrobots PTAL.

ofrobots:
Lgtm

PR #3, opened 2016-02-02
Update profile collection to match new api

matthewloring:


matthewloring:
@ofrobots ping. Putting this through for now to make progress. LGTM.

PR #5, opened 2016-02-23
Remove unsampling logic from agent

matthewloring:
Unsampling has been moved into V8 with
https://codereview.chromium.org/1706343002 so this logic is no longer
necessary in the agent.

This change also updates the units used for emitted profiles to be more
consistent with other profilers.


matthewloring:
@ofrobots PTAL.

ofrobots:
lgtm.

PR #7, opened 2017-09-25
Added code which captures how to start agent

nolanmar511:
Code here generally captures how the interface for starting the profiling agent should work.

Generally, one should be able to start the profiler by either including, then starting the profiler
`
var profiler = require('./index');
profiler.start(config); // where config implements Config interface
`

or by requiring the profiler at runtime
`node --require ./index some_program.js`

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -1,0 +1,8 @@
+export interface Config {
+  service?: string;
+  serviceVersion?: string;
+  projectId?: string;
+  instance?: string;
+  zone?: string;
+  apiAddr?: string;  
+}
--- a/index.ts
+++ b/index.ts
@@ -1,0 +1,27 @@
+'use strict'
+
+import { Config } from './config'
+
+/**
+* Start the profiling agent
+* Currently unimplemented
+*
+* @param {object=} config - Profiler configuration
+*
+* @example
+* profiler.start();
+*/
+function start(projectConfig: Config) {
+  console.log("UNIMPLEMENTED");
+}
+
+module.exports = {
+  start: start 
+};
+
+// If the module was --require'd from the command line, start the agent
+if (module.parent && module.parent.id === 'internal/preload') {
+  module.exports.start();
+}
+
+export default {};
--- a/index_test.js
+++ b/index_test.js
@@ -1,0 +1,4 @@
+var profiler = require('./index');
+
+profiler.start({});
+
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "protobufjs": "~5.0.1"
   },
   "devDependencies": {
+    "@types/node": "^8.0.30",
     "clang-format": "^1.0.33"
   }
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,9 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "lib": [ "es6" ],
+    "declaration": true,
+    "strict": true
+  }
+}


config.ts
```diff
@@ -0,0 +1,8 @@
+export interface Config {
```
aalexand: Document the fields. Probably just along the lines of https://code.googlesource.com/gocloud/+/master/profiler/profiler.go.

config.ts
```diff
@@ -0,0 +1,8 @@
...
+  service?: string;
+  serviceVersion?: string;
+  projectId?: string;
+  instance?: string;
+  zone?: string;
+  apiAddr?: string;  
+}
```
aalexand: Should we have a debugLogging parameter like Go agent does?

index.ts
```diff
@@ -0,0 +1,27 @@
...
+* Start the profiling agent
+* Currently unimplemented
+*
+* @param {object=} config - Profiler configuration
+*
+* @example
+* profiler.start();
+*/
+function start(projectConfig: Config) {
+  console.log("UNIMPLEMENTED");
```
aalexand: Is using `console.log` how people log things in Node.js? I'd expect there is something more specialized.

index.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+module.exports = {
+  start: start 
+};
+
+// If the module was --require'd from the command line, start the agent
+if (module.parent && module.parent.id === 'internal/preload') {
```
aalexand: This looks magic to me. Is this common to have such check in Node.js modules?

index.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+* Start the profiling agent
+* Currently unimplemented
+*
+* @param {object=} config - Profiler configuration
+*
+* @example
+* profiler.start();
+*/
+function start(projectConfig: Config) {
```
aalexand: Should this document how it signals an error and what may go wrong?

aalexand: `projectConfig` is rather just `config` or `agentConfig` imo.

index.ts
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+import { Config } from './config'
+
+/**
+* Start the profiling agent
```
aalexand: Nit: missing full stop. Please make all doc strings full sentences.

aalexand: Indentation on this and other lines below is wrong (missing one leading space).

config.ts
aalexand: Should we make the config as similar to Cloud Trace config as possible? See https://github.com/GoogleCloudPlatform/cloud-trace-nodejs. For example, that seems to use an environment variable to specify the project ID, and has the service parameter in a nested structure.

config.ts
```diff
@@ -0,0 +1,8 @@
+export interface Config {
+  service?: string;
```
ofrobots: Use a service context schema similar to this [definition](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/20e69f23d3da2f405f40cbe16df606159d199fa0/config.ts#L125) from trace agent (debug agent [definition](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/015f29c842111b96f6db062adee070f7bce83a5c/src/agent/config.ts#L34)).

Side-bar: we will get these into a common module down the road.

index.ts
```diff
@@ -0,0 +1,27 @@
...
+* Currently unimplemented
+*
+* @param {object=} config - Profiler configuration
```
ofrobots: JSDoc type annotations are not necessary anymore. Omit.

index.ts
nolanmar511: This is what cloud-trace does to allow the module to be started when required from the cli
(https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/index.ts#L179-L182).
I'm not sure how common this type of check would be, though.

cloud-debug doesn't seem to support users requiring it at run time.

nolanmar511:
I'm going to close this, and resubmit something which more closely resembles the config set-up in cloud-debug-nodejs

PR #11, opened 2017-10-09
add local CPU profile collection

nolanmar511:


src/profilers/cpu-profiler.ts
```diff
@@ -0,0 +1,43 @@
+import {perftools} from '../profile';
+import {serializeCpuProfile} from './profile-serializer';
+const cpuProfiler = require('bindings')('cpu_profiler');
+
+export class CpuProfiler {
```
nolanmar511: Not sure if this would be better as a namespace than as a class.

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,145 @@
+import {perftools} from '../profile';
```
ofrobots: add missing copyright

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,145 @@
...
+      timeNanos: 1000 * 1000 * prof.endTime,                         // Nanos
+      durationNanos: 1000 * 1000 * (prof.endTime - prof.startTime),  // Nanos
+
+      periodType: timeValueType,
+      period: this.samplingInterval
+    };
+  }
+
+  // clears fields used to build a profile.
+  private reset() {
```
ofrobots: Looking at the implementation, it seems that the objects of this class are meant to be instantiated only for the duration of the `serializeWallProfile` function, after which the object, and state it holds are no longer necessary.

The pattern of holding state in a temporary short lived object, along with a `reset` function here doesn't seem like the best idea to me.

Instead, I think it would make sense for `serializeWallProfile` to be a top-level function that creates a closure with the necessary state. All the other private functions in the class can be nested functions on `serializeWallProfile`, hence getting access to the closure state. WDYT?

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
+import {perftools} from '../profile';
```
ofrobots: copyright.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+const wallProfiler = require('bindings')('wall_profiler');
+
+export class WallProfiler {
+  // Average time in microseconds between samples.
+  samplingInterval: number;
+
+  // True when the profiler is actively profiling.
+  private profiling: boolean;
+
+  constructor(samplingInterval: number) {
```
ofrobots: Define this way: `constructor(private samplingInterval: number)` and get rid of property declaration above and assignment below.

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /* Adds samples from a node and it's children to the fields tracking
+   * profile serialization.
+   *
+   * node - the node which is serialized
+   * stack - the stack trace to the current node.
+   */
+  private serializeNode(node: WallProfileNode, stack: Stack) {
+    let that = this;
```
ofrobots: Change the `function(child)` below to a fat arrow function (`(child) => { ... }`), and this would not be needed anymore. Arrow functions keep the lexical this.

Even better, use a for..of loop instead of the `forEach`. `for (let child of node.children) {... }`. This avoids the extra closure allocation / function dispatch per iteration.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.profiling = false;
+    wallProfiler.setSamplingInterval(this.samplingInterval);
+  }
+
+  // Collects a profile for the duration, in milliseconds, specified by
+  // profileDuration.
+  // Returns a promise which will resolve to the collected profile.
+  // If profiling is already started, the returned promise will be rejected.
+  async profile(profileDuration: number): Promise<perftools.profiles.IProfile> {
+    const that = this;
```
ofrobots: Using arrow functions would avoid the need for this.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!that.profiling) {
```
ofrobots: Flip the condition and do an error return early. It makes the code more readable (no negation in the condition, error conditions are dealt with upfront, decreases the code indent for most of the function body.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+      return new Promise(function(resolve) {
+        setTimeout(function() {
+          let result = wallProfiler.stopProfiling('');
+          that.profiling = false;
+          let profile = serializeWallProfile(result, that.samplingInterval);
+          resolve(profile);
+        }, profileDuration);
+      });
+    }
+    return Promise.reject(new Error('already profiling with WallProfiler'));
```
ofrobots: You should throw the error instead of wrapping it into a rejected promise.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  constructor(samplingInterval: number) {
+    this.samplingInterval = samplingInterval;  // us
+    this.profiling = false;
+    wallProfiler.setSamplingInterval(this.samplingInterval);
+  }
+
+  // Collects a profile for the duration, in milliseconds, specified by
+  // profileDuration.
+  // Returns a promise which will resolve to the collected profile.
```
ofrobots: async functions always return a promise, so this description is overly wordy. `Returns the .. profile ..` would be appropriate wording here IMO.

src/profilers/wall-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+  // If profiling is already started, the returned promise will be rejected.
+  async profile(profileDuration: number): Promise<perftools.profiles.IProfile> {
+    const that = this;
+    if (!that.profiling) {
+      that.profiling = true;
+      wallProfiler.startProfiling('', true);
+      return new Promise(function(resolve) {
+        setTimeout(function() {
```
ofrobots: Use existing npm modules for async delays this rather than manually putting this together with Promises and setTimeout. This [delay](https://npm.im/delay) module looks nice.

test/test-profile-serializer.ts
```diff
@@ -0,0 +1,135 @@
+import {perftools} from '../src/profile';
```
ofrobots: copyright

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "build/src/index.js",
   "scripts": {
     "install": "node-gyp rebuild",
-    "test": "true",
+    "test": "mocha build/test/test-*.js",
```
ofrobots: Any specific reason for picking mocha?

aalexand:
@ofrobots We could keep the "CPU profiler" terminology in the API here but we'll still need to send these profiles as of "wall" profiling type in the Cloud Profiler speak to be consistent with the data agents for other languages produce. Those use "CPU" profiling type to describe profiles where samples represent the actual CPU time spent by the program.

src/profilers/profile-serializer.ts
nolanmar511: That makes more sense to me, too. Will change.

package.json
nolanmar511: No -- just saw that trace and debug seemed to use mocha and decided to be consistent. 
I'd be happy to change if you'd recommend something else.

src/profilers/wall-profiler.ts
```diff
@@ -1,39 +1,37 @@
...
-          resolve(profile);
-        }, profileDuration);
-      });
+    if (this.profiling) {
+      throw new Error('already profiling with WallProfiler');
     }
-    return Promise.reject(new Error('already profiling with WallProfiler'));
+    this.profiling = true;
+    wallProfiler.startProfiling('', true);
+    return delay(profileDuration).then(() => {
```
ofrobots:
````
wallProfiler.startProfiling('', true);  
await delay(profileDuration);
let result = wallProfiler.stopProfiling('');
// ...
```

nolanmar511:
PTAL 
(still trying to figure out why gts check fails, which is why travis fails)

package.json
ofrobots: mocha is what a lot of repos use indeed, and consistency is important. In gts itself we started using [ava](https://npm.im/ava) because it comes with good typescript and async function support out of the box. It also has some other nice features (parallel, better assertions).

However, at this point I don't think we have enough experience with ava to recommend it over mocha; but I would strongly encourage you to take a look at it.

One thing you may not necessarily like however, is that it runs tests in parallel. This means that your tests have to be non-interfering. While this seems a bit painful to get started with, I think this has longer term benefits in forcing healthier unit tests.

test/test-profile-serializer.ts
ofrobots: This still needs to be addressed.

test/test-profile-serializer.ts
```diff
@@ -0,0 +1,135 @@
...
+        'function2', 'script2', 'function5', 'script3', 'function4', 'function3'
+      ],
+      timeNanos: 100 * 1000 * 1000,
+      durationNanos: 100 * 1000 * 1000,
+      periodType: new perftools.profiles.ValueType({type: 3, unit: 4}),
+      period: 1000,
+    };
+
+    const timeProfileOut = serializeTimeProfile(timeProfile, 1000);
+    assert.deepEqual(timeProfileOut, expTimeProfileOut);
```
ofrobots: It would be good to add code coverage at this point as well. As reference, here's a PR that adds coverage to another repo: https://github.com/google/ts-style/commit/69db7abd00b7ed5ae6b46987b11edb25c6f7f5d8.

test/test-profile-serializer.ts
```diff
@@ -0,0 +1,135 @@
+import {perftools} from '../src/profile';
+import {serializeTimeProfile} from '../src/profilers/profile-serializer';
+import {TimeProfile, TimeProfileNode} from '../src/v8-types';
+
+let assert = require('assert');
+
+describe('serializeTimeProfile', function() {
```
ofrobots: here and elsewhere, prefer arrow functions.

test/test-time-profiler.ts
```diff
@@ -0,0 +1,89 @@
+import * as delay from 'delay';
```
ofrobots: copyright

test/test-time-profiler.ts
```diff
@@ -0,0 +1,89 @@
...
+let assert = require('assert');
+
+describe('TimeProfiler', function() {
+  describe('profile', function() {
+    it('should profile during duration and finish profiling after duration',
+       function() {
+         const duration = 500;
+         const sampleInterval = 1000;
+         let profiler = new TimeProfiler(sampleInterval);
+         let profilePromise = profiler.profile(duration);
```
ofrobots: These tests actually enable/disable the profiler, so these would be hard to get running in parallel.

Can you look into whether you can either dependency inject or mock the time profiler bindings so that these become true unit tests?

I think you would still want tests that actually enable profiling, but I would consider them 'system tests'. The unit tests should focus on unit behaviors.

test/test-time-profiler.ts
```diff
@@ -0,0 +1,89 @@
...
+         return delay(2 * duration).then(() => {
```
ofrobots: Making the test function async and then use `await` instead of `then`.

Here and elsewhere.

test/test-time-profiler.ts
```diff
@@ -0,0 +1,89 @@
...
+         const duration = 500;
+         const sampleInterval = 1000;
+         let profiler = new TimeProfiler(sampleInterval);
+         let profilePromise = profiler.profile(duration);
+         profilePromise.then(function(profile) {
+           if (profile.sampleType !== undefined) {
+             const vt1: perftools.profiles.IValueType = profile.sampleType[0];
+             const vt2: perftools.profiles.IValueType = profile.sampleType[1];
+
+             if (profile.stringTable !== undefined) {
```
ofrobots: These branches should be reversed (error condition first)

test/test-time-profiler.ts
```diff
@@ -0,0 +1,89 @@
...
+        if (profile.sample !== undefined) {
+          assert.ok(profile.sample.length > 0, 'there are no samples');
+        } else {
+          assert.fail('sample field of profile is undefined');
+        }
+        done();
+      });
+    });
+
+    it('should return rejected promise when already profiling', function(done) {
```
ofrobots: Use async/await. In which case `await profiler.profile()` would throw.

package.json
nolanmar511: This is more related to #12, but would if be possible still to change set value of process.env.GCLOUD_PROJECT (for example) in some tests and delete the value in other tests if I switched to ava?

I realize that running unit tests in parallel is ideal, but I also feel that it is very important to be able to modify the value of process.env.GCLOUD_PROJECT (and similarly used environment variables) in unit tests for initConfig().

nolanmar511:
PTAL

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,127 @@
...
+    for (let child of node.children) {
+      serializeNode(child, stack);
+    }
+    stack.shift();  // remove leaf from stack
+  }
+
+  function getLocation(node: TimeProfileNode): perftools.profiles.Location {
+    const id = node.callUid;
+    if (locationMap.has(id)) {
+      return locationMap.get(id) as perftools.profiles.Location;
```
aalexand: Instead of doing a cast, can you do get() and check whether the insertion is needed using that? It would also avoid doing two lookups on the common path of having the location present.

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,127 @@
...
+      samples.push(sample);
+    }
+    for (let child of node.children) {
+      serializeNode(child, stack);
+    }
+    stack.shift();  // remove leaf from stack
+  }
+
+  function getLocation(node: TimeProfileNode): perftools.profiles.Location {
+    const id = node.callUid;
```
aalexand: Curious: what is callUid? How is this assigned?

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+import {serializeTimeProfile} from './profile-serializer';
+
+const timeProfiler = require('bindings')('time_profiler');
+
+export class TimeProfiler {
+  // True when the profiler is actively profiling.
+  private profiling: boolean;
+
+  // samplingInterval is average time in microseconds between samples.
+  constructor(private samplingInterval: number) {
```
aalexand: samplingInterval -> intervalMicros. Please always include the abbreviation of the time unit when the type is a generic type like "number" (rather than something specific like "Duration"). E.g. fooMicros, fooNanos and so on.

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.profiling = false;
+    timeProfiler.setSamplingInterval(this.samplingInterval);
+  }
+
+  // Collects a profile for the duration, in milliseconds, specified by
+  // profileDuration.
+  // Returns the collected profile.
+  // If profiling is already started, throws error.
+  async profile(profileDuration: number): Promise<perftools.profiles.IProfile> {
```
aalexand: durationMillis

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (this.profiling) {
+      throw new Error('already profiling with TimeProfiler');
```
aalexand: already profiling with time profiler

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+    timeProfiler.startProfiling('', true);
+    await delay(profileDuration);
+    let result = timeProfiler.stopProfiling('');
+    this.profiling = false;
+    return serializeTimeProfile(result, this.samplingInterval);
+  }
+
+  // Returns true if the TimeProfiler is currently profiling and false
+  // otherwise.
+  isRunning(): boolean {
```
aalexand: Nit: isProfiling maybe? To be consistent with how the member variable is named. Or name that "running".

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+    this.profiling = true;
+    timeProfiler.startProfiling('', true);
+    await delay(profileDuration);
+    let result = timeProfiler.stopProfiling('');
+    this.profiling = false;
+    return serializeTimeProfile(result, this.samplingInterval);
+  }
+
+  // Returns true if the TimeProfiler is currently profiling and false
```
aalexand: Nit: TimeProfiler -> profiler. No need to repeat "TimeProfiler" again and again, it's clear from the context.

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,127 @@
...
+      systemName: name,
+      filename: getIndexOrAdd(node.scriptResourceName || '(unknown)', strings)
+      // start_line
+    });
+    functions.push(f);
+    functionMap.set(id, f);
+    return f;
+  }
+
+  function getSampleValueType(): perftools.profiles.ValueType {
```
aalexand: This and the function below seems more of "create" or "make" than "get".

bindings/time-profiler.cc
```diff
@@ -19,7 +19,7 @@
 
 using namespace v8;
 
-Local<Value> TranslateCpuProfileNode(const CpuProfileNode* node) {
+Local<Value> TranslateTimeProfileNode(const CpuProfileNode* node) {
   // TODO: Implement unimplemented interface
```
aalexand: This (and below) comment is not super-clear: which interface? Could you clarify?

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Collects a profile for the duration, in milliseconds, specified by
+  // profileDuration.
+  // Returns the collected profile.
+  // If profiling is already started, throws error.
+  async profile(profileDuration: number): Promise<perftools.profiles.IProfile> {
+    if (this.profiling) {
+      throw new Error('already profiling with TimeProfiler');
+    }
+    this.profiling = true;
+    timeProfiler.startProfiling('', true);
```
aalexand: Can multiple profiling sessions co-exist in parallel? What happens if user enables the profiling at the same time?  Is the string parameter here the way to arbitrage the concurrent collections? If so, maybe it would be better to generate a unique profile name?

bindings/time-profiler.cc
nolanmar511: @ofrobots -- I think this comment is from the original code. Any ideas what it refers to?

src/profilers/profile-serializer.ts
nolanmar511: @ofrobots -- I'm not certain what callUid is. Could you clarify?

bindings/time-profiler.cc
ofrobots: I am not quite sure. Feel free to remove it.

src/profilers/profile-serializer.ts
ofrobots: This term originates in the V8 code base. The documentation is sparse, but the comment describes it as 'function entry UID'. See: https://cs.chromium.org/chromium/src/v8/include/v8-profiler.h?q=v8-profiler&sq=package:chromium&dr&l=203

Looking at the source, it seems that this V8 API is deprecated anyway. Let's start using GetScriptId, GetLineNumber, and GetColumnNumber instead. (A follow-on PR is fine; make sure to open an issue for this however).

nolanmar511:
PTAL

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Adds samples from a node and it's children to the fields tracking
+   * profile serialization.
+   *
+   * @param node - the node which is serialized
+   * @param stack - the stack trace to the current node.
+   */
+  function serializeNode(root: TimeProfileNode) {
+    let nodes = [root];
```
ofrobots: Since you always push/pop nodes and stacks together, this suggests that you you should probably define a `Entry` data structure with a node and a stack. Push and pop atomic entities of this type.

src/profilers/profile-serializer.ts
```diff
@@ -64,12 +64,16 @@ export function serializeTimeProfile(
    * @param stack - the stack trace to the current node.
    */
   function serializeNode(root: TimeProfileNode) {
-    let nodes = [root];
-    let stacks: number[][] = [[]];
-    while (nodes.length > 0) {
-      let node = nodes.pop();
-      let stack = stacks.pop();
-      if (node !== undefined && stack !== undefined) {
+    let entries: {node: TimeProfileNode, stack: number[]}[] = [];
```
ofrobots: nit: declaring this as an interface rather than an inline type definition would be better.

nolanmar511:
PTAL

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (f !== undefined) {
+      return f;
+    }
+    const name = getIndexOrAdd(node.functionName || '(anonymous)', strings);
+    f = new perftools.profiles.Function({
+      id: id,
+      name: name,
+      systemName: name,
+      filename: getIndexOrAdd(node.scriptResourceName || '(unknown)', strings)
+      // start_line
```
aalexand: Elaborate or remove.

src/profilers/profile-serializer.ts
```diff
@@ -0,0 +1,146 @@
...
+        {functionId: getFunction(node).id, line: node.lineNumber});
+  }
+
+  function getFunction(node: TimeProfileNode): perftools.profiles.Function {
+    const id = node.callUid;
+    let f = functionMap.get(id);
+    if (f !== undefined) {
+      return f;
+    }
+    const name = getIndexOrAdd(node.functionName || '(anonymous)', strings);
```
aalexand: getIndexOrAdd is O(n), which makes the profile conversion quadratic the number of functions plus the number of source files. I think we need a more efficient lookup.

src/profilers/time-profiler.ts
```diff
@@ -0,0 +1,44 @@
...
+    profiler.setSamplingInterval(this.intervalMicros);
+  }
+
+  /**
+   * Collects a profile for the duration specified.
+   *
+   * @param durationMillis - time in milliseconds for which to collect profile.
+   */
+  async profile(durationMillis: number): Promise<perftools.profiles.IProfile> {
+    const runName = 'stackdriver-profiler-' + Date.now() + '-' + Math.random();
```
aalexand: Remove "stackdriver-" from the prefix, I think we should limit using the product name to the documentation as much as possible.

src/v8-types.ts
```diff
@@ -16,21 +16,21 @@
 
 // Type Definitions based on implementation in bindings/
 
-export interface CpuProfile {
+export interface TimeProfile {
   /** Time in nanoseconds at which profile was stopped. */
   endTime: number;
```
aalexand: endTimeNanos

src/v8-types.ts
```diff
@@ -16,21 +16,21 @@
...
-  topDownRoot: CpuProfileNode;
+  topDownRoot: TimeProfileNode;
   /** Time in nanoseconds at which profile was started. */
   startTime: number;
```
aalexand: startTimeNanos

PR #12, opened 2017-10-13
feat: get instance and zone from metadata

nolanmar511:


src/config.ts
```diff
@@ -64,11 +64,14 @@ export interface Config extends AuthenticationConfig {
 
 // Default values for configuration for a profiler.
 export const defaultConfig: Config = {
-  projectId: '{{projectId}}',
   logLevel: 1,
+  projectId: '{{projectId}}',
   serviceContext: {
+    service: 'default',
```
aalexand: Like in Go agent, we should require setting the service name, it should not fall back to anything. Except on Flex where it can be derived from the environment so the user is not required to set it (but if they do, that will override the environment setting).

src/index.ts
```diff
@@ -13,49 +13,94 @@
...
 const common: Common = require('@google-cloud/common');
+const metadataAPI = 'http://metadata.google.internal/computeMetadata/v1/';
+
+// Returns a promise which will resolve to true if running on GCE and false
+// otherwise.
+// It is assumed a program is running on GCE if the metadata API can be
+// accessed.
+function onGCE(): Promise<boolean> {
+  return new Promise<boolean>((resolve) => {
+    request.get(
```
aalexand: Will it fail fast enough when running outside of GCE? Or do we need to set a short timeout?

Also, this function could probably later be made a part of some shared code.

src/index.ts
```diff
@@ -13,49 +13,94 @@
...
+        {url: metadataAPI, headers: {'Metadata-Flavor': 'Google'}},
+        function(err: Error, response) {
+          err ? resolve(false) : resolve(true);
+        });
+  });
+}
+
+// Returns a promise that resolves to the value of metadata field.
```
aalexand: As @ofrobots mentioned in some other PR, there isn't much need to describe the semantics of returning a promise over and over again as it's pretty idiomatic. I think the documentation here and in all other methods that return a promise can be simplified to something like "Returns a GCE metadata value or fails if there is problem communicating the metadata server or the HTTP request fails otherwise."

src/index.ts
```diff
@@ -13,49 +13,94 @@
...
+    request.get(
+        {url: metadataAPI + field, headers: {'Metadata-Flavor': 'Google'}},
+        function(err: Error, response: request.RequestResponse) {
+          err ? reject(err) : resolve(response.body);
+        });
+  });
+}
 
 // initConfig sets unset values in the configuration to the value retrieved from
 // environment variables, metadata, or the default value specified in
```
aalexand: default values

src/index.ts
```diff
@@ -13,49 +13,94 @@
...
-    logLevel: process.env.GCLOUD_PROFILER_LOGLEVEL,
+  const envConfig: any = {
     projectId: process.env.GCLOUD_PROJECT,
     serviceContext: {
       service: process.env.GAE_SERVICE,
       version: process.env.GAE_VERSION,
     }
   };
 
+  if (process.env.GCLOUD_PROFILER_LOGLEVEL !== undefined) {
```
aalexand: Do I read it correct that if I set GCLOUD_PROFILER_LOGLEVEL to something that is not a number (like "foo"), the logging will be enabled at a rather verbose level 10?

nolanmar511: The second argument means parseInt() will parse values to be in base 10.
If GCLOUD_PROFILER_LOGLEVEL is not set, or set to something that can't be parsed as an int, it will be ignored.

src/index.ts
nolanmar511: I've run this and it does return pretty quickly, but I'll specify a timeout of 2 seconds, since that seems to be the maximum time golang's version of this function could take. (https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/compute/metadata/metadata.go#L249-L251)

src/index.ts
```diff
@@ -13,49 +13,100 @@
...
-import {AuthenticationConfig, Common, ServiceConfig} from '../third_party/types/common-types';
+import * as request from 'request';
+
+import {AuthenticationConfig, Common, ServiceConfig, ServiceObject} from '../third_party/types/common-types';
+
 import {Config, defaultConfig} from './config';
 import {Profiler, ProfilerConfig} from './profiler';
 
 const common: Common = require('@google-cloud/common');
+const metadataAPI = 'http://metadata.google.internal/computeMetadata/v1/';
```
ofrobots: Any reason you're not using the [gcp-metadata](https://npm.im/gcp-metadata) module and querying the metadata service manually?

src/index.ts
```diff
@@ -13,49 +13,100 @@
...
+        },
+        function(err: Error, response) {
+          err ? resolve(false) : resolve(true);
+        });
+  });
+}
+
+// Returns value of metadata field.
+// Throws error if there is a problem accessing metadata API.
+function getField(field: string): Promise<string> {
```
ofrobots: This functionality is provided by gcp-metadata.

src/index.ts
```diff
@@ -13,49 +13,100 @@
...
   }
 
-  let normalizedConfig = extend(true, {}, defaultConfig, envConfig, config);
+  let normalizedConfig =
+      extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
-  if (normalizedConfig.serviceContext.service === undefined) {
-    throw new Error('service name must be specified in the configuration');
+  // fetch instance and zone from metadata if not specified and on GCE
+  if (await onGCE()) {
```
ofrobots: Reverse the order of these ifs. If you have a locally provided instance/zone, you don't need to await `onGCE` (i.e. no need to add a network latency when you don't need it). 

Secondly, I don't think you really need to check if you're on GCE before attempting to fetch instance/zone. You should fetch these values (if not provided by user) from the metadata service blindly, and if you're not on GCE the requests will fail.

Thirdly, you should fetch instance and zone in parallel rather than serializing those latencies. Something like:

```
const [instance, zone] = await Promsie.all([getField('instance/name'), getField('instance/zone')]);
```

src/index.ts
```diff
@@ -13,49 +13,100 @@
...
+      });
+      normalizedConfig.zone = zone.substring(zone.lastIndexOf('/') + 1);
+    }
   }
 
-  // TODO: fetch instance and zone from metadata. This will require function to
-  // be asynchrous.
-  if (normalizedConfig.instance === undefined) {
-    normalizedConfig.instance = '';
+  if (normalizedConfig.projectId === undefined) {
```
ofrobots: Requiring the projectId from the user is too aggressive compared to the other agents. The authentication flow will automatically figure out the project id (or give an error) if one is not provided.

test/test-init-config.ts
```diff
@@ -0,0 +1,359 @@
+import * as assert from 'assert';
```
ofrobots: copyright

src/index.ts
nolanmar511: I generally agree with your comment, but I am a little concerned about changing this to what you've suggested, because I believe we want to require that the instance and zone are specified when code is being run from GCE, but not require that these fields are specified off of GCE. 

I'm currently echoing the pattern used by the go profiling agent (https://code.googlesource.com/gocloud/+/master/profiler/profiler.go#363) 

@aalexand -- thoughts?

ofrobots: > we want to require that the instance and zone are specified when code is being run from GCE, but not require that these fields are specified off of GCE.

My suggestion doesn't contradict this. I am suggesting that that the 'onGCE' check is not necessary and you get the exact same outcome regardless.

nolanmar511: I was worried about the situation in which instance or zone might not be set in the metadata.
But I think this should be okay, given that it does look like these fields are set in metadata.

src/index.ts
nolanmar511: I'm a bit concerned about how slow gcp-metadata is. 
It took about  6 seconds until it threw an error when I ran it locally, and I don't see a clear way to make it faster.

src/index.ts
nolanmar511: I tried out gcp-metadata, and I found it took about 6 seconds to throw an error when I ran it locally.
There didn't seem to be a way to adjust the query/retries in gcp-metadata to make it faster, and 6 seconds felt like a long time to delay profiling when one is running outside of GCE.

nolanmar511:
PTAL

src/index.ts
ofrobots: That's unexpected and should be investigated. If you have a minimal repro, can you open a bug against gcp-metadata?

src/config.ts
```diff
@@ -64,11 +64,13 @@ export interface Config extends AuthenticationConfig {
 
 // Default values for configuration for a profiler.
 export const defaultConfig: Config = {
-  projectId: '{{projectId}}',
   logLevel: 1,
+  projectId: '{{projectId}}',
```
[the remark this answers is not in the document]
nolanmar511: The projectId is a part of the url to communicate with the profiler API (for example, https://cloudprofiler.googleapis.com/v2/projects/{{projectId}}/profiles),  so it needs to be set.
Can this still be omitted? Can we trust this to be filled in correctly?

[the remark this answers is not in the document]
ofrobots: That's the case for [trace as well](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects/patchTraces). This particular string is an implementation detail of the @ google-cloud/common repo, and is likely to be changed going forward. I would like to avoid promulgating it elsewhere. Please avoid if at all possible.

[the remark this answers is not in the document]
ofrobots: Perhaps the `projectIdRequired` property on `ServiceAuthenticationConfig` is a way to avoid specifying this. Can you take a look?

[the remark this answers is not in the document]
nolanmar511: That should work.

nolanmar511:
PTAL

test/test-init-config.ts
```diff
@@ -0,0 +1,336 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as assert from 'assert';
+import * as nock from 'nock';
+import * as request from 'request';
```
ofrobots: This is unused. Please drop the dependency on this from `package.json`.

package.json
```diff
@@ -21,22 +21,31 @@
...
   "dependencies": {
+    "@google-cloud/common": "^0.13.6",
     "bindings": "^1.2.1",
     "delay": "^2.0.0",
     "extend": "^3.0.1",
-    "nan": "^2.1.0"
+    "gcp-metadata": "^0.4.0",
+    "nan": "^2.1.0",
+    "request": "^2.83.0",
+    "retry-request": "^3.0.1"
```
ofrobots: this is unused.

package.json
```diff
@@ -21,22 +21,31 @@
...
   "devDependencies": {
+    "@types/assert": "0.0.31",
     "@types/delay": "^2.0.0",
     "@types/extend": "^2.0.30",
     "@types/mocha": "^2.2.43",
+    "@types/nock": "^8.2.1",
     "@types/node": "^8.0.30",
+    "@types/request": "^2.0.4",
     "@types/sinon": "^2.3.6",
+    "assert": "^1.4.1",
```
ofrobots: assert is a built-in module. You don't need it in the package.json.

package.json
```diff
@@ -21,22 +21,31 @@
...
     "delay": "^2.0.0",
     "extend": "^3.0.1",
-    "nan": "^2.1.0"
+    "gcp-metadata": "^0.4.0",
+    "nan": "^2.1.0",
+    "request": "^2.83.0",
+    "retry-request": "^3.0.1"
   },
   "devDependencies": {
+    "@types/assert": "0.0.31",
```
ofrobots: @types/node should already contain these types.

src/config.ts
```diff
@@ -64,11 +64,17 @@ export interface Config extends AuthenticationConfig {
...
     version: '',
   },
   disableHeap: false,
   disableTime: false,
+  instance: '',
+  zone: ''
+};
+
+// Configuration parameters set internally, not by users.
+export const internalConfig = {
```
ofrobots: You don't need this here. Simply add it to the Service object config here: https://github.com/nolanmar511/cloud-profiler-nodejs/blob/41428d20b54112e67c39888ceef56f5732f89506/src/profiler.ts#L46

src/index.ts
```diff
@@ -13,49 +13,80 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import * as extend from 'extend';
 import * as path from 'path';
-import {AuthenticationConfig, Common, ServiceConfig} from '../third_party/types/common-types';
-import {Config, defaultConfig} from './config';
+const gcpMetadata = require('gcp-metadata');
```
ofrobots: Can you use import?

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
+
+import {AuthenticationConfig, Common, ServiceConfig, ServiceObject} from '../third_party/types/common-types';
+
+import {Config, defaultConfig, internalConfig} from './config';
 import {Profiler, ProfilerConfig} from './profiler';
 
 const common: Common = require('@google-cloud/common');
+const metadataAPI = 'http://metadata.google.internal/computeMetadata/v1/';
```
ofrobots: This is unused.

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
+
+// Returns value of metadata field.
+// Throws error if there is a problem accessing metadata API.
+function getField(field: string): Promise<string> {
```
ofrobots: You can replace this function by using`pify` and its `multiArgs` option: https://www.npmjs.com/package/pify#multiargs

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
-// Returns rejected promise if value that must be set cannot be initialized.
-async function initConfig(config: Config): Promise<ProfilerConfig> {
+// Throws error if value that must be set cannot be initialized.
+// Exported for testing purposes.
+export async function initConfig(config: Config): Promise<ProfilerConfig> {
   config = common.util.normalizeArguments(null, config);
 
-  const envConfig = {
-    logLevel: process.env.GCLOUD_PROFILER_LOGLEVEL,
+  const envConfig: any = {
```
ofrobots: Can we do better than `any` here? Unless I am mistaken, all the properties in `Config` are optional, so this can at least use the `Config` type?

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
 
-  // TODO: fetch instance and zone from metadata. This will require function to
-  // be asynchrous.
-  if (normalizedConfig.instance === undefined) {
-    normalizedConfig.instance = '';
-  }
-  if (normalizedConfig.zone === undefined) {
-    normalizedConfig.zone = '';
+  if (normalizedConfig.serviceContext.service === undefined) {
+    throw new Error('service must be specified in the configuration');
```
ofrobots: Is it possible to auto-detect the service and throw only in the case where you cannot possibly identify the serivce? Put differently, if we cannot detect the service, it is almost always our rather than the user's fault.

See the trace agent for how we deal with this.

I'd be okay if you want to do this in a follow-up, but please open an issue in that case.

test/test-init-config.ts
```diff
@@ -0,0 +1,336 @@
...
+import * as sinon from 'sinon';
+
+import {initConfig} from '../src/index';
+
+const gcpMetadata = require('gcp-metadata');
+
+
+nock.disableNetConnect();
+const metadataAPI = 'http://metadata.google.internal/computeMetadata/v1';
```
ofrobots: This is unused. Did you run lint?

test/test-init-config.ts
```diff
@@ -0,0 +1,336 @@
...
+ */
+
+import * as assert from 'assert';
+import * as nock from 'nock';
+import * as request from 'request';
+import * as sinon from 'sinon';
+
+import {initConfig} from '../src/index';
+
+const gcpMetadata = require('gcp-metadata');
```
ofrobots: use import.

test/test-init-config.ts
```diff
@@ -0,0 +1,336 @@
...
+
+describe('initConfig', () => {
+  let processLogLevel: string|undefined;
+  let processCloudProject: string|undefined;
+  let processService: string|undefined;
+  let processVersion: string|undefined;
+  let processConfig: string|undefined;
+
+  before(() => {
+    processLogLevel = process.env.GCLOUD_PROFILER_LOGLEVEL;
```
ofrobots: How about something like

```
savedEnv = process.env;
process.env = {};

// later in after()
process.env = savedEnv;
```

test/testdata/test-config.json
```diff
@@ -0,0 +1,12 @@
+{
```
ofrobots: nit: we prefer the term `fixtures` over `testdata`.

src/index.ts
nolanmar511: I don't think so -- gcpMetadata doesn't have type definitions.

src/index.ts
aalexand: For App Engine flexible environment we should use GAE_SERVICE and GAE_VERSION for the service and service version default values respectively, similar to how the Go profiling agent does.

src/index.ts
ofrobots: But it does, as of very recently: https://github.com/stephenplusplus/gcp-metadata/pull/9

src/index.ts
nolanmar511: We currently do use GAE_SERVICE and GAE_VERSION (in envConfig), but if GAE_SERVICE is not set, then service might not be set, and an error is thrown.
I believe this is how the go profiling agent works.

test/test-init-config.ts
nolanmar511: I've run gts check and gts fix -- am I missing something?

src/index.ts
nolanmar511: Okay -- I think I can now that I've updated the version used.

src/index.ts
```diff
@@ -15,26 +15,25 @@
...
 // Returns value of metadata field.
 // Throws error if there is a problem accessing metadata API.
 function getField(field: string): Promise<string> {
-  return new Promise<string>((resolve, reject) => {
-    gcpMetadata.instance(
-        field, (err: Error, response: any, metadata: string) => {
-          err ? reject(err) : resolve(metadata);
-        });
-  });
+  return pify(gcpMetadata.instance, {multiArgs: true})(field).then(
```
ofrobots: I think you can store the pify return value at the top-level.

```
const metadataInstance = pify(gcpMetadata.instance, {multiArgs: true});
```
I wonder if there would be much left in this function after that, but I'll let you decide.

ofrobots: If you decide to keep this function, use a different make, e.g. `getMetadataInstanceField`.

nolanmar511:
PTAL

test/test-init-config.ts
ofrobots: Hmm.. that should have caught it.. if not, that sounds like a bug in gts.

src/index.ts
```diff
@@ -28,7 +28,7 @@ const common: Common = require('@google-cloud/common');
 
 // Returns value of metadata field.
 // Throws error if there is a problem accessing metadata API.
-function getField(field: string): Promise<string> {
+function getMetadataInstanceField(field: string): Promise<string> {
```
ofrobots: nit: make this an async function and `await` instead of `.then`.

nolanmar511:
@aalexand - any comments?

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
+  }
+
   let envSetConfig: Config = {};
   if (process.env.hasOwnProperty('GCLOUD_PROFILER_CONFIG')) {
     envSetConfig =
         require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
   }
 
-  let normalizedConfig = extend(true, {}, defaultConfig, envConfig, config);
+  let normalizedConfig =
```
aalexand: Can this be named just `config` or maybe `mergedConfig`?  I don't think "normalized" is a good term here.

src/index.ts
```diff
@@ -13,49 +13,80 @@
...
 
-  // TODO: fetch instance and zone from metadata. This will require function to
-  // be asynchrous.
-  if (normalizedConfig.instance === undefined) {
-    normalizedConfig.instance = '';
-  }
-  if (normalizedConfig.zone === undefined) {
-    normalizedConfig.zone = '';
+  if (normalizedConfig.serviceContext.service === undefined) {
+    throw new Error('service must be specified in the configuration');
```
aalexand: Is having the error message start with lowercase and not have a period at the end intentional? I just don't know what's the best style for these but make sure it's consistent with other places of reporting an error.

nolanmar511: I looked at trace -- seems like capitalization with period is used there. 
I've fix this here.

src/index.ts
nolanmar511: Done

nolanmar511:
Could someone merge this in for me?
I would like to do the fixes needed before we can merge in #26 on top of this change, if possible, because those changes will require a lot of changes to files which this also touched.

aalexand:
@ofrobots Can you merge? I don't have the Merge button enabled.

ofrobots:
Merge is disabled because the CI/checks are not green. Coverage dropped more than an acceptable amount.

ofrobots:
I landed this. I couldn't see anything pressing in the coverage report that should block this PR at this point.

PR #16, opened 2017-10-20
feat: add communications with Stackdriver Profiler API

nolanmar511:


src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
 
-// TODO: finish implementing Profiler.
-// TODO: add stop() method to stop profiling.
+// Converts a profile to a compressed, base64 encoded string.
+function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
+  let time = new Date((p.timeNanos as number) / 1000000);
+  let pwriter = perftools.profiles.Profile.encode(p);
+  let buffer = new Buffer(pwriter.finish());
+  return new Promise<string>((resolve, reject) => {
+    zlib.gzip(buffer, (err: Error, gzBuf: Buffer) => {
```
ofrobots: It would be a lot cleaner if you used `util.promisify` or `pify` to promisify this API and then simply awaited `zlib.gzip`. You'll need to make this function async.

There are a few other examples in this PR that would benefit from the same pattern.

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
+      }
+    }
+  }
+
+  // Talks to Stackdriver Profiler server to create profile.
+  // If any problem is encountered, an error will be thrown.
+  // TODO: retry rather than fail when createProfile() throws error.
+  // Public to allow for testing.
+  async createProfile(): Promise<RequestProfile> {
+    let reqBody = {
```
ofrobots: Use `const` instead of `let` unless you need a `let`.

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
+  // Starts and endless loop to poll Stackdriver Profiler server for
+  // instructions, and collects and uploads profiles as requested.
+  // If there is a problem when collecting a profile or uploading a profile to
+  // Stackdriver Profiler, this problem will be logged at the debug level.
+  // If there is a problem polling Stackdriver Profiler for instructions
+  // on the type of profile created, an error will be thrown.
   async start(): Promise<void> {
-    return Promise.reject(
-        new Error('start() is unimplemented for ProfileAgent'));
+    while (true) {
```
ofrobots: This would keep a program alive forever.

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
+  // Public to allow for testing.
+  async profileAndUpload(prof: RequestProfile): Promise<void> {
+    prof = await this.writeTimeProfile(prof);
+    let options = {
+      method: 'PATCH',
+      uri: API + '/' + prof.name,
+      body: prof,
+      json: true,
+    };
+    let serviceObject = this.serviceObject;
```
ofrobots: Why do you need this variable?

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+import * as delay from 'delay';
```
ofrobots: Is this used?

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
+  let buffer = new Buffer(pwriter.finish());
+  return new Promise<string>((resolve, reject) => {
+    zlib.gzip(buffer, (err: Error, gzBuf: Buffer) => {
+      err ? reject(err) : resolve(gzBuf.toString('base64'));
+    });
+  });
+}
+
+// Service object for communicating with Stackdriver Profiler API.
+class ProfilerServiceObject extends common.ServiceObject {
```
ofrobots: Did you consider making `Profiler` a subclass of `common.ServiceObject`. This class has nothing but a constructor and is supposed to be an exclusive private property of `Profiler`.

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
+  }
+}
+
+// Polls Stackdriver Profiler server for instructions on behalf of a task and
+// collects and uploads profiles as requested.
 export class Profiler {
-  config: ProfilerConfig;
-  logger: Logger;
-  service: Service;
+  private config: ProfilerConfig;
```
ofrobots: While you're making this change; remove this from here, and instead mark it as a private property in the constructor argument list. This will also get rid of need for the `this.config = ` assignment.

src/profiler.ts
```diff
@@ -13,57 +13,203 @@
...
     this.logger = new common.logger({
       level: common.logger.LEVELS[config.logLevel as number],
       tag: pjson.name
     });
+
+    // TODO: enable heap profiling once heap-profiler implemented.
+    this.profileTypes = [];
+    if (!this.config.disableTime) {
+      this.profileTypes.push('WALL');
+      this.timeProfiler = new TimeProfiler(1000);
```
ofrobots: Declare a constant variable at the top or in the class instead of providing it here.

Secondly, should this be configurable by the user and exposed in the config instead?

src/profiler.ts
nolanmar511: Yeah -- I wasn't certain as to a way around this. I know we need to ask the API what profile we should collect, then collect and upload the profile.

I'd considered modifying it so it would be like 
 ` ` ` 
async start(): Promise<void> {
   this.continueProfiling = true;
   while(this.continueProfiling) {
      // stuff
  }
 // more stuff
}

stop() {
  this.continueProfiling = false;
}
 ` ` ` 

But that would really only help when a user only had a few copies of the same projectId/service running, since most time is spent in createProfile() waiting for the API to indicate one should profile. And since the API only profiles once a minute per projectId/service/version combination, this wouldn't be very helpful when there are lots of jobs with the same projectId/service/version.

src/profiler.ts
nolanmar511: I was able to use pify here, but I couldn't figure out how to get this to work for serviceObject.request()  -- I got errors at run time related to serviceObject being undefined. Could you point me to any examples for promisifying a method?

src/profiler.ts
nolanmar511: I think we need to decide what characteristics of profiling should be configurable and then make those things configurable.
I'll file an issue, and that can be a separate PR going forward.

nolanmar511:
PTAL

src/profiler.ts
```diff
@@ -13,57 +13,228 @@
...
-  instance: string;
-  zone: string;
-  disableTime: boolean;
-  disableHeap: boolean;
+const API = 'https://cloudprofiler.googleapis.com/v2';
+const gzip = pify(zlib.gzip);
+
+const WALL_TYPE = 'WALL';
+const HEAP_TYPE = 'HEAP';
+const SAMPLING_INTERVAL_MICROS = 1000;
```
ofrobots: Whether or not this is user exposed, please avoid hard-coded constants like this and instead put this in `config` (or possibly `config.internal`).

nolanmar511:
PTAL

src/profiler.ts
```diff
@@ -13,57 +13,222 @@
...
+   * instructions, and collects and uploads profiles as requested.
+   * If there is a problem when collecting a profile or uploading a profile to
+   * Stackdriver Profiler, this problem will be logged at the debug level.
+   * If there is a problem polling Stackdriver Profiler for instructions
+   * on the type of profile created, an error will be thrown.
+   */
   async start(): Promise<void> {
-    return Promise.reject(
-        new Error('start() is unimplemented for ProfileAgent'));
+    while (true) {
```
ofrobots: As discussed in person, we need to get rid of the this infinite loop as it will keep the application alive forever.

Instead we should pick a minimum interval (say `minInterval`) that we want to allow between profiles and setTimeout/delay for `minInterval - time taken by createProfile`. The server controls the max delay between profiles, but the agent should have a notion of the minimum delay.

Secondly we need a comment before the call to createProfile to indicate that it uses a hanging `GET`. Looking at the code it seems that it has "only" the latency of a normal http GET, whereas in reality it a lot longer.

src/profiler.ts
```diff
@@ -13,57 +13,222 @@
...
+const HEAP_TYPE = 'HEAP';
+
+/**
+ * Interface for body of response from Stackdriver Profiler API when creating
+ * profile and used as body of request to Stackdriver Profiler API when
+ * uploading a profile.
+ *
+ * Public for testing.
+ */
+export interface RequestProfile {
```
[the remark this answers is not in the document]
nolanmar511: This definition is basically equivalent to this: https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudprofiler/v2/profiler.proto#L87.

But since we're using REST rather than gRPC, we aren't currently using profiler.proto.

src/profiler.ts
```diff
@@ -13,57 +13,222 @@
...
+  private logger: Logger;
+  private profileTypes: string[];
+  private continueProfiling: boolean;
+
+  // Public for testing.
+  timeProfiler: TimeProfiler|undefined;
 
   constructor(config: ProfilerConfig) {
-    this.config = config;
+    config = common.util.normalizeArguments(null, config);
```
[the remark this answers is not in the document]
nolanmar511: When I tried this, I see the error "'super' must be called before accessing 'this' in the constructor of a derived class."

src/profiler.ts
```diff
@@ -13,57 +13,222 @@
...
+      } catch (err) {
+        this.logger.debug(err);
+      }
+    }
+  }
+
+  /**
+   * Talks to Stackdriver Profiler server to create profile.
+   * If any problem is encountered, an error will be thrown.
+   * TODO: retry rather than fail when createProfile() throws error.
```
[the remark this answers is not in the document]
nolanmar511: Sorry -- that TODO probably isn't clear enough. 
Our agents generally have retries with exponential back-off (up to one hour) for profile creation (the response can also direct the back-off). I'll clarify this.

src/profiler.ts
```diff
@@ -13,57 +13,222 @@
...
+      profileType: this.profileTypes,
+    };
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+    return new Promise<any>((resolve, reject) => {
+      this.request(
```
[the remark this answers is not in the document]
nolanmar511: How would I use pify for this.

I tried:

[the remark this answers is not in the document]
ofrobots: It turns out that `this.request` is already promisified. See: https://github.com/GoogleCloudPlatform/google-cloud-node/blob/941c79245d79943c7226430ec543ac225579c00f/packages/common/src/service-object.js#L386.

src/profiler.ts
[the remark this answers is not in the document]
ofrobots: Ack.

src/profiler.ts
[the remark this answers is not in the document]
ofrobots: But given that `this.request` returns multiple args, I suspect that the promisification done by `util.promisifyAll` is buggy. @stephenplusplus do you expect `request` to work promisifed?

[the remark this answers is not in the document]
stephenplusplus: Pinging @callmehiphop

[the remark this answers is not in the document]
callmehiphop: If you want to use pify you need to bind a context

```js
return pify(this.request.bind(this), { multiArgs: true })(options)
  .then(([body, response]) => {});
```

As far as promisifying `request`, we made a decision to always return an Array, even if there's only a single item being resolved, so I believe this to be functioning as expected. However it's unlikely that anyone would ever call it directly.

nolanmar511:
PTAL

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+
+  // Public for testing.
+  timeProfiler: TimeProfiler|undefined;
 
   constructor(config: ProfilerConfig) {
-    this.config = config;
+    config = common.util.normalizeArguments(null, config);
+    const serviceConfig = {
+      baseUrl: API,
+      scopes: ['https://www.googleapis.com/auth/monitoring.write'],
```
ofrobots: Is this the right scope for Cloud Profiler? From the name, it sounds like this scope would be for Stackdriver Monitoring API.

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+   * instructions, and collects and uploads profiles as requested.
+   * If there is a problem when collecting a profile or uploading a profile to
+   * Stackdriver Profiler, this problem will be logged at the debug level.
+   * If there is a problem polling Stackdriver Profiler for instructions
+   * on the type of profile created, an error will be thrown.
+   */
   async start(): Promise<void> {
-    return Promise.reject(
-        new Error('start() is unimplemented for ProfileAgent'));
+    return this.createAndUploadProfile(0);
```
ofrobots: Use default parameters in `createAndUploadProfile` and omit the 0 from here.

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+
+  /**
+   * Talks to Stackdriver Profiler server to create profile when instructed
+   * and then upload that profile.
+   *
+   * @param delayMillis - milliseconds to wait before polling Stackdriver
+   * Profiler server to see if it's time to create a profile and profile.
+   */
+  async createAndUploadProfile(delayMillis: number): Promise<void> {
+    this.profilingTimeoutId = setTimeout(async () => {
```
ofrobots: Why is this assignment (and property) necessary? I don't see use in any other method besides this one.

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+      const startCreateMillis = Date.now();
+      const prof = await this.createProfile();
+      await this.profileAndUpload(prof);
+      const endCreateMillis = Date.now();
+      this.createAndUploadProfile(
+          this.config.minTimeBetweenProfilesMillis -
+          (endCreateMillis - startCreateMillis));
+    }, delayMillis);
+    this.profilingTimeoutId.unref();
+  }
```
ofrobots: An alternative way to write this function would be:

```js
async createAndUploadProfile() {
  const startCreateMillis = Date.now();
  prof = await this.createProfile();
  const endCreateMillis = Date.now();

  // Schedule the next profile.
  const delay = this.config.minTimeBetweenProfilesMillis - (endCreateMillis - startCreateMillis);
  setTimeout(this.createAndUploadProfile.bind(this), delay).unref();
}
```

But don't change code to this just yet.

Other things to think about:
* createProfile and profileAndUpload are called back to back. Do they need to be separate functions?
* While setTimeout correctly deals with negative timeouts, it would be good to clamp the value at 0 on the low end.
* This function is creating, uploading and scheduling, but the name only describes the first two verbs. Also note that start is a single line function that does a single opeartion: 'create and upload profile', but in reality it is running and async loop.
* Another way to fix this might be to split this function into two logical parts: the scheduling logic and the profiling logic. If you decide to follow this, perhaps your original structure of the could is better compared to the inverted form I gave earlier in this comment.

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+      profileType: this.profileTypes,
+    };
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+
+    return this.request(options).then((result) => {
```
ofrobots: You're in an async function, so you can `await` rather than `then`ing. Rule of thumb: There should be very few (if any) reasons to use `.then` in this codebase given that we can use `await` now.

src/profiler.ts
```diff
@@ -13,57 +13,236 @@
...
+  async profileAndUpload(prof: RequestProfile): Promise<void> {
+    prof = await this.writeTimeProfile(prof);
+    const options = {
+      method: 'PATCH',
+      uri: API + '/' + prof.name,
+      body: prof,
+      json: true,
+    };
+
+    return this.request(options).then((result) => {
```
ofrobots: ditto.

test/test-profiler.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const API = 'https://cloudprofiler.googleapis.com/v2';
+
+const mockTimeProfiler = mock(TimeProfiler);
+when(mockTimeProfiler.profile(10 * 1000)).thenReturn(new Promise((resolve) => {
+  resolve(testProfile);
+}));
+
+nock.disableNetConnect();
+export function oauth2(): nock.Scope {
```
ofrobots: * This needs a better name given that it is supposed to be used directly now. `nockOauth2` would be better.
* This does not need to be exported.

src/config.ts
```diff
@@ -62,6 +61,19 @@ export interface Config extends AuthenticationConfig {
...
+  zone: string;
+  disableTime: boolean;
+  disableHeap: boolean;
+  timeSamplingIntervalMicros: number;
+  minTimeBetweenProfilesMillis: number;
+}
+
 // Default values for configuration for a profiler.
 export const defaultConfig: Config = {
   projectId: '{{projectId}}',
```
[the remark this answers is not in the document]
nolanmar511: I've updated this in #12. I will make sure that this string is not included when I merge whichever is approved first into the other.

src/profiler.ts
nolanmar511: I believe so -- this is the scope used by other profiling agents.

nolanmar511:
PTAL

src/profiler.ts
```diff
@@ -121,17 +120,19 @@ export class Profiler extends common.ServiceObject {
...
+  async coordinateProfiling(delayMillis = 0): Promise<void> {
+    const startCreateMillis = Date.now();
+    const prof = await this.createProfile();
+    await this.profileAndUpload(prof);
+    const endCreateMillis = Date.now();
+
+    // Schedule the next profile.
+    let delay = this.config.minTimeBetweenProfilesMillis -
+        (endCreateMillis - startCreateMillis);
+    if (delay < 0) {
```
ofrobots: [Math.max](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max)

src/profiler.ts
```diff
@@ -121,17 +120,19 @@ export class Profiler extends common.ServiceObject {
...
-      const startCreateMillis = Date.now();
-      const prof = await this.createProfile();
-      await this.profileAndUpload(prof);
-      const endCreateMillis = Date.now();
-      this.createAndUploadProfile(
-          this.config.minTimeBetweenProfilesMillis -
-          (endCreateMillis - startCreateMillis));
-    }, delayMillis);
-    this.profilingTimeoutId.unref();
+  async coordinateProfiling(delayMillis = 0): Promise<void> {
```
ofrobots: You don't use the parameter anymore. You will also need to update the JSDoc which no longer reflect what the function does. IMO, the name is not very intuitive either.

src/profiler.ts
```diff
@@ -186,9 +185,7 @@ export class Profiler extends common.ServiceObject {
       json: true,
     };
 
-    return this.request(options).then((result) => {
-      return;
-    });
+    await this.request(options);
```
ofrobots: What's the intention around how errors in the request will be dealt with?

src/profiler.ts
```diff
@@ -162,10 +163,8 @@ export class Profiler extends common.ServiceObject {
       json: true,
     };
 
-    return this.request(options).then((result) => {
-      const [body, response] = result;
-      return body;
-    });
+    const [body, response] = await this.request(options);
```
ofrobots: What's the intention around how errors in the request will be dealt with?

src/profiler.ts
nolanmar511: Other profilers generally ignore errors which occur when attempting to upload the profile. I intend to do the same.

nolanmar511: Errors in collecting or uploading the profile will be ignored. 
I'll add this now, since it doesn't really complicate this too much.

src/profiler.ts
nolanmar511: Here, an error should be thrown for certain types of errors (I need to check what we do with our other profilers to be certain which types of errors) which we know won't be resolved going forward. Otherwise, profile creation should be retried with exponential back-off, capped at one hour. 
I'll implement the logic for backing off on retries and checking which errors should be retried in a later PR.

src/profiler.ts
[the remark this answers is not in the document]
nolanmar511: I think it's worth considering, but it does have a few imports. So (I think) to be able to compile it to javascript, we would need to add more than  just `profiler.proto`.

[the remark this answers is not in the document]
ofrobots: It is okay to consider this out of scope for this PR, but I would still like to see this happen. Do you have the full list of protos we need? Presumably gRPC based profiler agents already have this full list in their repos?

src/profiler.ts
```diff
@@ -13,57 +13,251 @@
...
+    }
+    const options = {
+      method: 'PATCH',
+      uri: API + '/' + prof.name,
+      body: prof,
+      json: true,
+    };
+    try {
+      await this.request(options);
+    } catch (err) {
```
ofrobots: See comment above about network vs. http errors.

src/profiler.ts
```diff
@@ -13,57 +13,251 @@
...
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+    try {
+      const [body, response] = await this.request(options);
+      return body;
+    } catch (err) {
```
ofrobots: Note that there are two types of errors possible here. Network errors (which will be thrown) and HTTP errors, which will be indicated by a non-200 HTTP statusCode in the response. The code as written, ignores HTTP errors, and assumes all network errors are ignorable.

Some more questions you need to ask: are there a subset of errors that are fatal enough that you should terminate (e.g. a ENOTFOUND thrown, or a HTTP 404 response), or if the right thing to do is to always ignore such errors and keep retrying the same operation (which are likely to fail the same way).

src/profiler.ts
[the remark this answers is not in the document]
aalexand: I agree we should do this, I'll comment on #27 with some details.

src/profiler.ts
aalexand: It is the right scope, we currently reuse it.

nolanmar511:
PTAL

src/profiler.ts
```diff
@@ -13,57 +13,292 @@
...
+   * minTimeBetweenProfilesMillis.
+   */
+  async pollProfilerService(): Promise<void> {
+    const startCreateMillis = Date.now();
+    const prof = await this.createProfile();
+    await this.profileAndUpload(prof);
+    const endCreateMillis = Date.now();
+
+    // Schedule the next profile.
+    const delayMillis = Math.max(
```
aalexand: We shouldn't have this delay. The profiles are collected at exponential distribution points with the mean of 1 minute, capping the min will skew the mean.

src/profiler.ts
```diff
@@ -13,57 +13,292 @@
...
+      return body;
+    } catch (err) {
+      retryRequest = isRetriableError(err);
+      requestError = new Error('Error creating profile: ' + err);
+      this.logger.debug(requestError);
+    }
+    if (retryRequest) {
+      // TODO: check response to see if response specifies a backoff.
+      // TODO: implement exponential backoff.
+      await delay(this.config.minTimeBetweenProfilesMillis);
```
aalexand: I understand we don't have the exponential backoff yet, that could be a separate PR, but please name that variable like something with "backoff" in its name. Look at the Go profiling agent code maybe.

src/profiler.ts
ofrobots: The concern I have here is that a misbehaving server, proxy or network can cause `createProfile` to return early, causing the agent to spin the CPU on the node process. Given that we're talking about a single thread here, this can have a pretty major impact on the user's application.

We need to defend against this possibility. Is there no way to avoid the skew?

aalexand: Errors like that should result in an HTTP / response error which are handled and back off. It might make sense to validate that the response from the server has the unique profile ID set.

ofrobots: I'm okay with any solution that guarantees that we won't spin the CPU.

nolanmar511:
PTAL

ofrobots:
Can you rebase, so that another build gets triggered? As it is, the coverage seems to drop substantially.

src/profiler.ts
```diff
@@ -33,36 +40,267 @@ export interface ProfilerConfig extends AuthenticationConfig {
...
+   * uploads profiles as requested.
+   */
+  async pollProfilerService(): Promise<void> {
+    const startCreateMillis = Date.now();
+    const prof = await this.createProfile();
+    await this.profileAndUpload(prof);
+    const endCreateMillis = Date.now();
+
+    // Schedule the next profile.
+    setTimeout(this.pollProfilerService.bind(this), 0).unref();
```
ofrobots: Use `setImmediate` when the timeout is 0.

src/profiler.ts
```diff
@@ -33,36 +40,267 @@ export interface ProfilerConfig extends AuthenticationConfig {
...
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+
+    let requestError: Error|undefined = undefined;
+    let retryRequest = true;
+    try {
+      const [body, response] = await this.request(options);
```
ofrobots: `this.request` already includes `retry-request` logic. Is that not sufficient?

nolanmar511: Since we want to keep retrying indefinitely in certain cases, I don't think the logic in retry-request will be sufficient.

ofrobots: In that case, I think you need higher level retry logic. Looking at HTTP response code and figuring out when to retry is duplicating logic from `retry-request`. Over time, this will drift apart from what `retry-logic` does.

nolanmar511:
@ofrobots -- I've been getting this:
```
----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
All files |  Unknown |  Unknown |  Unknown |  Unknown |                |
----------|----------|----------|----------|----------|----------------|
```
from code coverage when I run it locally. Code coverage doesn't seem to be being updated.  I think this was just since I merged in changes this morning.

Any ideas how to fix this?

aalexand:
@nolanmar511 @ofrobots There was a question on how CreateProfile call (which is hanging) can be cancelled, as I think unless it is cancelled the Node.js program being profiled won't exit? Looking at the code of google-cloud-node/common, some thoughts:

* `makeAuthenticatedRequestFactory` in util.js does seem to support cancellation via abort() on the active request - see [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0434b4df9ebd001021cc16281981f69a0a764248/packages/common/src/util.js#L391).
* Unfortunately in the service.js file the return value seems to be [dropped](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0434b4df9ebd001021cc16281981f69a0a764248/packages/common/src/service.js#L174) as there isn't a return statement. I don't know why.
* It also dropped down the chain [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0434b4df9ebd001021cc16281981f69a0a764248/packages/common/src/service-object.js#L353) and [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0434b4df9ebd001021cc16281981f69a0a764248/packages/common/src/service-object.js#L366).
* Also, all methods on a service object are wrapped with `util.promisifyAll`, [here](https://github.com/GoogleCloudPlatform/google-cloud-node/blob/0434b4df9ebd001021cc16281981f69a0a764248/packages/common/src/service-object.js#L386). The wrapper only returns a promise if no callbacks are passed though - if there are callbacks passed, it returns the original value.

So, I would naively expect that supporting cancellation of CreateProfile call is as easy as:
* Change the code of service.js and service-object.js to return the object with the abort() call for non-streaming requests.
* Change the profiler agent call to pass callbacks for the CreateProfile's request() call to promisify the call manually to gain the control over abort() call and call that on program exit.
* Not sure how to intercept into the program exit. There is `process.on('exit', ...)` but not sure if there is a chicken-and-egg issue with it.

ofrobots:
@nolanmar511 fix for coverage issue here: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/pull/33. My recent changes regressed this.

aalexand:
One possible solution

ofrobots:
@aalexand did you close this by mistake?

aalexand:
@ofrobots Yes, sorry, re-opened.

nolanmar511:
PTAL
I am still planning on switching to using type definitions from profiler.proto, implementing exponential back off, and getting back off from the server, but I think these things are best addressed in later PRs.
I did add heap profile collection to profiler.ts, because it was a small amount of code allowed me to write meaningful tests for functions that had previously just thrown errors.

ts/src/config.ts
```diff
@@ -61,25 +61,35 @@ export interface Config extends AuthenticationConfig {
...
-  serviceContext: {service?: string; version?: string;};
+  serviceContext: {service: string; version?: string;};
   instance: string;
   zone: string;
   disableTime: boolean;
   disableHeap: boolean;
+  timeIntervalMicros: number;
+  heapIntervalBytes: number;
+  heapMaxStackDepth: number;
+  backoffMillis: number;
```
aalexand: I don't think we should expose it in the config.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
-// TODO: add stop() method to stop profiling.
-export class Profiler {
-  config: ProfilerConfig;
-  logger: Logger;
-  service: Service;
+/**
+ * Returns true if http status code indicates request should be retried.
+ */
+function isRetriableResponseCode(code: number) {
+  // TODO: determine which codes one should not retry on.
```
aalexand: Retry the same things the Go agent retries?

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+ * uploading a profile.
+ *
+ * Public for testing.
+ */
+export interface RequestProfile {
+  name?: string;
+  profileType?: string;
+  duration?: any;
+  profileBytes?: string;
+  labels?: {instance?: string; zone?: string};
```
aalexand: zone should only be a deployment label - https://code.googlesource.com/gocloud/+/master/profiler/profiler.go#397

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+}
+
+/**
+ * Converts a profile to a compressed, base64 encoded string.
+ *
+ * @param p - profile to be converted to string.
+ */
+async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
```
aalexand: Please confirm (and probably document) that nothing in this function is supposed to block the node thread. The call pwriter.finish() looked a bit suspicious to me - like as if blocks until the encoding operation finishes. Probably that's not true. In general I assume nothing in the agent blocks the node thread.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+    return this.pollProfilerService();
+  }
+
+  /**
+   * Endlessly polls the profiler server for instructions, and collects and
+   * uploads profiles as requested.
+   */
+  async pollProfilerService(): Promise<void> {
+    const startCreateMillis = Date.now();
+    const prof = await this.createProfile();
```
aalexand: Did we figure out the cancellation of this call? If not, add a TODO and describe the current plan in the PR description?

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+    await this.profileAndUpload(prof);
+    const endCreateMillis = Date.now();
+
+    // Schedule the next profile.
+    setImmediate(this.pollProfilerService.bind(this)).unref();
+  }
+
+  /**
+   * Talks to Stackdriver Profiler server, which hangs until server indicates
```
aalexand: "Talks to the profiler server, ...". As I asked before, please limit usage of the full product name to just once in the module (or in a source file) to avoid the repetitions.

ts/src/config.ts
```diff
@@ -61,25 +61,35 @@ export interface Config extends AuthenticationConfig {
...
   instance: '',
   zone: ''
 };
+
+// Configuration parameters set internally, not by users.
+export const internalConfig = {
+  timeIntervalMicros: 1000,
+  heapIntervalBytes: 512 * 1024,
+  heapMaxStackDepth: 64,
+  backoffMillis: 1000,  // 1 second
```
aalexand: One second backoff without exponential increase is way too low.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+   * retry createProfile.
+   *
+   * Public to allow for testing.
+   */
+  async createProfile(): Promise<RequestProfile> {
+    const reqBody = {
+      deployment: {
+        projectId: this.config.projectId,
+        target: this.config.serviceContext.service,
+        labels: {zone: this.config.zone, instance: this.config.instance},
```
aalexand: The deployment labels should not include the instance. The point of having deployments as a separate thing is to have constant profiling rate among number of instances.

You do need to pass the version as a deployment label, you don't seem to pass anywhere?

Read the source code of Go the agent to ensure such things are consistent between the two.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+      profileType: this.profileTypes,
+    };
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+
+    let requestError: Error|undefined = undefined;
```
aalexand: Nit: requestError -> error or reqErr? Any particular reason for longer name?

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+    let retryRequest = true;
```
aalexand: Nit: just 'retry'?

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
-// TODO: finish implementing Profiler.
-// TODO: add stop() method to stop profiling.
-export class Profiler {
-  config: ProfilerConfig;
-  logger: Logger;
-  service: Service;
+/**
+ * Returns true if http status code indicates request should be retried.
+ */
+function isRetriableResponseCode(code: number) {
```
aalexand: isRetriableStatusCode or isRetriableResponseStatus (saying 'status' is important since that's the error domain here)

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+      requestError = new Error('Error creating profile: ' + err.toString());
+      this.logger.debug(requestError);
+    }
+    if (retryRequest) {
+      // TODO: check response to see if response specifies a backoff.
+      // TODO: implement exponential backoff.
+      await delay(this.config.backoffMillis);
+      return this.createProfile();
+    }
+    throw requestError;
```
aalexand: This should never happen unless the agent has a bug. Log.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+  }
+
+  /**
+   * Collects a profile of the type specified by the profileType field of prof.
+   * If any problem is encountered, like a problem collecting or uploading the
+   * profile, an error will be logged at the debug level, but otherwise ignored.
+   *
+   * Public to allow for testing.
+   *
+   * @param prof
```
aalexand: What's the point of this "@param prof" markup I am curious? It does not explain anything, does not document the type, and I can clearly see that the function has one parameter named `prof` from the source.

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+            new Error('Error creating profile: ' + response.statusMessage);
+        this.logger.debug(requestError);
+      } else {
+        return body;
+      }
+    } catch (err) {
+      retryRequest =
+          isRetriableError(err) || isRetriableResponseCode(err.statusCode);
+      requestError = new Error('Error creating profile: ' + err.toString());
+      this.logger.debug(requestError);
```
aalexand: Why logged as a debug level and not an error?

ts/src/profiler.ts
```diff
@@ -13,47 +13,292 @@
...
+      default:
+        throw new Error('Unexpected profile type ' + prof.profileType + '.');
+    }
+  }
+
+  /**
+   * Collects a time profile, converts profile to compressed, base64 encoded
+   * string, and adds profileBytes field to prof with this string.
+   *
+   * Public to allow for testing.
```
aalexand: I don't see this function being tested, is the comment out of date? Can you check other functions? I would also prefer not to have these individual functions tested as much as possible anyway and rather have the test exercise the overall behavior talking to some kind of fake profile server so that things are close to what they are going to be in production.

ts/src/config.ts
nolanmar511: Right now, users cannot set this value -- it would be overridden by the value in internalConfig.
Is there a better way to do this?

ts/src/config.ts
nolanmar511: Done -- increasing backoff to 5 minutes.

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: The statuses we'd retry on would be different, because go grpc seems to return codes between 0 and 14 (inclusive) (https://godoc.org/google.golang.org/grpc/codes), not http status codes.  So, I need to figure out how the status codes relate.

ts/src/profiler.ts
nolanmar511: Added a TODO referencing issue for this.

ts/src/profiler.ts

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: Switched to logging at error level -- that seems to be what is done for profiler.go

ts/src/profiler.ts
nolanmar511: Done. I removed this.

ts/src/profiler.ts
nolanmar511: This function (and other functions with this in the comment) are tested. 
(https://github.com/nolanmar511/cloud-profiler-nodejs/blob/bb77e78eb4250b5694762af286284e5be4c51771/ts/test/test-profiler.ts#L131).

cloud-trace-nodejs and cloud-debug-nodejs have unit tests (like the ones I've written), and system tests (like what you've described) -- I know I need to add system tests, but should I also get rid of unit tests?

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: Done -- now match go code.

ts/src/profiler.ts
nolanmar511: @aalexand -- what should I do if the response status code (or error thrown by the http request) indicates we should not retry? That's the only case in which this code would be executed.

nolanmar511:
I've realized that the projectId is used in the body of the request for profile creation. So, I think the projectId needs to be retrieved from metadata (with an error thrown when it wasn't specified and cannot be retrieved).
I have made this change here, but mention it because I know throwing an error when projectId is not specified has been debated.

ts/src/profiler.ts
aalexand: Not retrying the profile creation means the profiling loop should exit, right? Under which conditions you would want to do so? It is safer to keep retrying any error once an hour (max cap under the randomization and backoff exponential increase that you should implement soon) than making a good decision that some errors should permanently stop the agent loop. So, which errors would you anticipate as being that much permanent?

ts/src/profiler.ts
aalexand: Ah, OK, keep then.

ts/src/profiler.ts
aalexand: Ping

ts/src/profiler.ts
aalexand: Ping

nolanmar511: Still looking into pwriter.finish(), but I think the gzBuf.toString() is problematic. 
I think I need to re-write this function.

nolanmar511: I've looked into this.
perftools.profiles.Profile.encode(p) is the part of this function that takes a majority of the time -- for a very large profile (cppbench, I merged different profiles from this, and ended up with a profile with a total duration of about 4 hours), and perftools.profiles.Profile.encode(p) blocked the event loop for like 500ms. 

I'm not sure if this can be avoid. Before I look more into it, is this definitely a problem?

nolanmar511: After meeting yesterday, I think any stalling will be short (and unavoidable).

nolanmar511:
PTAL

ts/src/config.ts
```diff
@@ -61,25 +61,41 @@ export interface Config extends AuthenticationConfig {
...
   serviceContext: {},
   disableHeap: false,
   disableTime: false,
   instance: '',
   zone: ''
 };
+
+// Configuration parameters set internally.
+// One must be very careful if modifying these values. Modifying these values
+// is not reccomended.
```
aalexand: recommended

ts/src/config.ts
```diff
@@ -61,25 +61,41 @@ export interface Config extends AuthenticationConfig {
...
   logLevel: number;
-  serviceContext: {service?: string; version?: string;};
+  serviceContext: {service: string; version?: string;};
   instance: string;
   zone: string;
   disableTime: boolean;
   disableHeap: boolean;
+
+  // Internal configuration parameters.
+  // One must be very careful if modifying values for these fields.
```
aalexand: Nit: one of the two sentences "must be very careful" and "not recommended" is redundant.

ts/src/config.ts
```diff
@@ -61,25 +61,41 @@ export interface Config extends AuthenticationConfig {
...
+  // Modifying values for these fields is not reccomended.
```
aalexand: recommended

ts/src/profiler.ts
```diff
@@ -13,47 +13,323 @@
...
+ */
+function isErrorResponseStatusCode(code: number) {
+  return code < 200 || code >= 300;
+}
+
+/**
+ * Returns true if http status code indicates request should be retried.
+ */
+function isRetriableResponseStatusCode(code: number) {
+  // TODO: determine which codes one should not retry on.
```
aalexand: For this and the isRetriableError functions, I feel we should either figure the conditions in this pull request, or remove the associated logic and just always retry for now. Having these functions doesn't serve much purpose right now, and I would be worried that the code stays this way for long time and it's a confusing state.

ts/src/config.ts

ts/src/config.ts

ts/src/config.ts

ts/src/profiler.ts
nolanmar511: Removing these for now.

nolanmar511:
PTAL

ts/src/config.ts
```diff
@@ -61,25 +61,39 @@ export interface Config extends AuthenticationConfig {
...
   disableHeap: false,
   disableTime: false,
   instance: '',
   zone: ''
 };
+
+// Configuration parameters set internally.
+// Modifying these values is not recommended.
+export const internalConfig = {
+  timeIntervalMicros: 1000,
```
aalexand: Let's change this to 10000 so that the frequency is 100 Hz like in Java and Go CPU profilers. Technically the Node.js' profiler is wall time profiler, but 100 Hz should be still good.

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+
+    try {
+      const results = await this.request(options);
+      // TODO: check types, don't cast.
```
aalexand: Why not do it now?

ts/src/config.ts
```diff
@@ -61,25 +61,39 @@ export interface Config extends AuthenticationConfig {
...
+  heapIntervalBytes: 512 * 1024,
```
aalexand: I would document the meaning of these parameters a bit more, so that it's clearer why changing them is not recommended and what are the tradeoffs of the default values.

ts/src/config.ts
```diff
@@ -61,25 +61,39 @@ export interface Config extends AuthenticationConfig {
...
   serviceContext: {},
   disableHeap: false,
   disableTime: false,
   instance: '',
   zone: ''
 };
+
+// Configuration parameters set internally.
+// Modifying these values is not recommended.
+export const internalConfig = {
```
ofrobots: I would suggest adding it to defaultConfig. That allows up to tweak/experiment without making source changes to the module. Everything in JS is public anyway, but this way is more convenient:

```ts
export const defaultConfig = {
  logLevel: 1,
  serviceContext: {},
  disableHeap: false,
  disableTime: false,
  instance: '',
  zone: '',
  internal: {
    timeIntervalMicros: 1000,
    heapIntervalBytes: 512 * 1024,
    heapMaxStackDepth: 64,
    backoffMillis: 5 * 60 * 1000,  // 5 minutes
  } 
};
```

This will also force you to declare types for these, and document them. That is a Good Thing even for internal configuration, IMO.

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+import * as http from 'http';
 import * as path from 'path';
+import * as pify from 'pify';
+import * as zlib from 'zlib';
 
-import {AuthenticationConfig, Common, Logger, Service, ServiceConfig} from '../third_party/types/common-types';
-
-import {Config, ProfilerConfig} from './config';
+import {perftools} from '../../proto/profile';
+import {AuthenticationConfig, Common, Logger, Service, ServiceConfig, ServiceObject, ServiceObjectConfig} from '../third_party/types/common-types';
```
ofrobots: overlong line. Use `gts fix`.

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+  /**
+   * Starts and endless loop to poll profiler server for
+   * instructions, and collects and uploads profiles as requested.
+   * If there is a problem when collecting a profile or uploading a profile to
+   * profiler server, this problem will be logged at the debug level.
+   * If there is a problem polling profiler server for instructions
+   * on the type of profile created, this problem will be logged. If the problem
+   * indicates one definitely will not be able to profile, an error will be
+   * thrown.
+   */
```
ofrobots: The comment should explain when the promise will be settled. The promise will settle once the profiler is started? Can it ever reject?

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
   async start(): Promise<void> {
```
ofrobots: You should omit `async` if you are not using `await` inside the function.

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+    };
+    const options = {
+      method: 'POST',
+      uri: '/profiles',
+      body: reqBody,
+      json: true,
+    };
+
+    try {
+      const results = await this.request(options);
```
ofrobots: Use destructuring assignment:
```ts
const [body, response] = await this.request(options);
```

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+      const body = results[0] as RequestProfile;
+      const response = results[1] as http.ServerResponse;
+
+      if (isErrorResponseStatusCode(response.statusCode)) {
+        this.logger.error('Error creating profile: ' + response.statusMessage);
+      } else {
+        return body;
+      }
+    } catch (err) {
+      this.logger.error('Error creating profile: ' + err.toString());
```
ofrobots:
```ts
this.logger.error(`Error creating profile: ${err}`);
```

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+    }
+
+    // TODO: determine which codes and errors one should not retry on.
+    // TODO: check response to see if response specifies a backoff.
+    // TODO: implement exponential backoff.
+    await delay(this.config.backoffMillis);
+    return this.createProfile();
```
ofrobots: Recursive call. This will stack overflow eventually. Also note that you are creating a chain of promises that never resolve.

ts/src/profiler.ts
```diff
@@ -13,47 +13,299 @@
...
+      return;
+    }
+    const options = {
+      method: 'PATCH',
+      uri: API + '/' + prof.name,
+      body: prof,
+      json: true,
+    };
+    try {
+      const results = await this.request(options);
```
ofrobots: Use destructuring assignment.

ts/src/profiler.ts
nolanmar511: I'll fix this, but thought I'd mention I just updated gts (and gts fix has caught other overlong lines),  but gts fix did not change this line.

nolanmar511: After fixing this, the gts check failed, so I left it as it is.

nolanmar511:
PTAL

ts/src/profiler.ts
```diff
@@ -13,47 +13,387 @@
...
+  Heap = 'HEAP'
+}
+
+/**
+ * Parses string containing a time duration, and returns this duration in ms.
+ * If duration cannot be parsed, returns undefined.
+ *
+ * Public for testing.
+ */
+export function parseDurationMillis(durationString: string): number|undefined {
```
ofrobots: nit: prefer `null` over `undefined` for the sentinel value. Generally I equate `undefined` with 'lack of value' and `null` with 'the nothing value'. Somewhat relevant, if a bit obtuse, [read](http://thecodelesscode.com/case/6).

ts/src/profiler.ts
```diff
@@ -13,47 +13,387 @@
...
+  if (!/(d+h)?(d+m)?(d+s)?(d+ms)?(d+us)?(d+ns)?/.test(durationString)) {
```
ofrobots: Do modules like [parse-duration](https://www.npmjs.com/package/parse-duration) help for this?

ts/src/profiler.ts
```diff
@@ -13,47 +13,387 @@
...
+   * Starts and endless loop to poll profiler server for
+   * instructions, and collects and uploads profiles as requested.
+   * If there is a problem when collecting a profile or uploading a profile to
+   * profiler server, this problem will be logged at the debug level and
+   * otherwise ignored.
+   * If there is a problem polling profiler server for instructions
+   * on the type of profile created, this problem will be logged and profile
+   * creation will be retried.
+   * This promise will not resolve while profiling is still occurring. This
+   * promise should not ever be resolved or rejected.
```
ofrobots: Let's not return a promise (or anything else) then.

The comment above needs to be updated. It no longer matches the implementation.

ts/src/profiler.ts
```diff
@@ -13,47 +13,387 @@
...
+    const delayMillis = await this.collectProfile();
+
+    // Schedule the next profile.
+    setTimeout(this.runLoop.bind(this), delayMillis).unref();
+  }
+
+  /**
+   * Waits for profiler server to tell it to collect a profile, then collects
+   * a profile and uploads it.
+   * Returns time, in ms, to wait before asking profiler server if one should
```
ofrobots: Use `@return` annotation rather than free form comment to describe the return value.

ts/src/profiler.ts
nolanmar511: That will work. Switching to parse-duration.

nolanmar511:
PTAL

ts/src/config.ts
aalexand: Done.

ts/src/config.ts
```diff
@@ -59,27 +59,54 @@ export interface Config extends AuthenticationConfig {
...
+  // reducing number of samples.
+  // Decreasing bytes between samples may increase overhead of profiling.
+  heapIntervalBytes?: number;
+
+  // Maximum depth of stacks recorded for heap samples. Decreasing stack depth
+  // will make it more likely that stack traces are truncated. Increasing
+  // stack depth may increase overhead of profiling.
+  heapMaxStackDepth?: number;
+
+  // Time to wait before trying to create a profile again if profile creation
```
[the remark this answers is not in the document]
nolanmar511: I'll skip this for now, but will use this level of detail when I add exponential back off.

ts/src/profiler.ts
```diff
@@ -13,47 +13,339 @@
...
+import {ProfilerConfig} from './config';
 import {HeapProfiler} from './profilers/heap-profiler';
 import {TimeProfiler} from './profilers/time-profiler';
 
+export const common: Common = require('@google-cloud/common');
+const parseDuration: (str: string) => number = require('parse-duration');
 const pjson = require('../../package.json');
-const common: Common = require('@google-cloud/common');
+const API = 'https://cloudprofiler.googleapis.com/v2';
+const scope = 'https://www.googleapis.com/auth/monitoring.write';
```
ofrobots: nit: use `SCOPE` in caps to indicate constant string (or be consistent, and convert `API` to lowercase).

ts/src/profiler.ts
```diff
@@ -13,47 +13,339 @@
...
+  async collectProfile(): Promise<number> {
+    let prof: RequestProfile;
+    try {
+      prof = await this.requestProfile();
+    } catch (err) {
+      this.logger.error(
+          `Error requesting profile type to be collected: ${err}`);
+      return this.config.backoffMillis;
+    }
+    await this.profileAndUpload(prof);
```
ofrobots: If `this.profileAndUpload` throws, would that cause the user application to terminate? That is, I don't see a catch block that would catch this error in the call chain.

ts/src/profiler.ts
```diff
@@ -13,47 +13,339 @@
...
+  }
+
+  /**
+   * Collects a time profile, converts profile to compressed, base64 encoded
+   * string, and puts this string in profileBytes field of prof.
+   *
+   * Public to allow for testing.
+   */
+  async writeTimeProfile(prof: RequestProfile): Promise<RequestProfile> {
+    if (this.timeProfiler) {
```
ofrobots: nit: reverse this if so that error is of the way early.

ts/src/profiler.ts
```diff
@@ -13,47 +13,339 @@
...
-    return Promise.reject(
-        new Error('start() is unimplemented for ProfileAgent.'));
+  /**
+   * Collects a heap profile, converts profile to compressed, base64 encoded
+   * string, and adds profileBytes field to prof with this string.
+   *
+   * Public to allow for testing.
+   */
+  async writeHeapProfile(prof: RequestProfile): Promise<RequestProfile> {
+    if (this.heapProfiler) {
```
ofrobots: ditto

ts/src/profiler.ts
nolanmar511: Errors are caught and logged in profileAndUpload().
I'll switch this, though, for consistency.

ts/src/profiler.ts

ts/src/profiler.ts

ts/src/profiler.ts

ts/src/profiler.ts

nolanmar511:
PTAL

nolanmar511:
@aalexand -- PTAL

ts/src/profiler.ts
```diff
@@ -13,47 +13,335 @@
...
+}
+
+/**
+ * Interface for body of response from profiler API when creating
+ * profile and used as body of request to profiler API when
+ * uploading a profile.
+ *
+ * Public for testing.
+ */
+export interface RequestProfile {
```
aalexand: If we add a new field in the profile structure, will parsing fail? Will the new field be lost between us getting the profile from profile creation and sending it to UpdateProfile?

ts/src/profiler.ts
```diff
@@ -13,47 +13,335 @@
...
+  profileBytes?: string;
+  deployment?: Deployment;
+  labels?: {instance: string};
+}
+
+/**
+ * @return true if an prof is a RequestProfile and false otherwise.
+ */
+// tslint:disable-next-line: no-any
+function isRequestProfile(prof: any): prof is RequestProfile {
```
aalexand: Why would this ever not be the case? E.g. why the successful response from CreateProfile not ever be a request profile?

ts/src/profiler.ts
aalexand: As we discussed, document this in the code.

ts/src/profiler.ts
```diff
@@ -13,47 +13,335 @@
...
+   */
+  start() {
+    this.runLoop();
+  }
+
+  /**
+   * Endlessly polls the profiler server for instructions, and collects and
+   * uploads profiles as requested.
+   */
+  async runLoop(): Promise<void> {
```
aalexand: start() function above does not use the return value of this runLoop() function - is this intentional?

ts/src/profiler.ts
```diff
@@ -13,47 +13,335 @@
...
+   * the specified type.
+   *
+   * TODO (issue #28): right now, this call could hang for up to an hour when
+   * this method is the only thing on the event loop, keeping the program open
+   * even when all work is done. Should expose the ability to cancel the http
+   * request made here, and then determine when to cancel this request.
+   *
+   * Public to allow for testing.
+   */
+  async requestProfile(): Promise<RequestProfile> {
```
aalexand: Call it `createProfile` to match the API call name.

ts/src/profiler.ts
```diff
@@ -13,47 +13,335 @@
...
+      throw new Error(
+          'Server response missing status information when attempting to upload profile.');
+    }
+    if (isErrorResponseStatusCode(response.statusCode)) {
+      let message: number|string = response.statusCode;
+      // tslint:disable-next-line: no-any
+      if ((response as any).statusMessage) {
+        message = response.statusMessage;
+      }
+      throw new Error(`Could not upload profile: ${message}`);
```
aalexand: Nit: full stop is missing. Can you go over other messages and check that all has the full stop?

ts/src/profiler.ts
nolanmar511: This is mostly just type-guarding, so that typescript knows that the object returned is a request profile.

ts/src/profiler.ts
nolanmar511: This is intentional. runLoop() only returns a promise because it is asychronous and so must return a promise.

aalexand: Maybe add a comment on that subtlety then.

ofrobots: I would recommend dropping the return type annotation instead.

nolanmar511: Thanks -- that works!

ts/src/profiler.ts
nolanmar511: This should not happen, and I added a test to createProfile() to confirm this.

ts/src/profiler.ts

ts/src/profiler.ts

nolanmar511:
PTAL

ts/src/profiler.ts
```diff
@@ -13,47 +13,358 @@
...
+   * Collects a time profile, converts profile to compressed, base64 encoded
+   * string, and puts this string in profileBytes field of prof.
+   *
+   * Public to allow for testing.
+   */
+  async writeTimeProfile(prof: RequestProfile): Promise<RequestProfile> {
+    if (!this.timeProfiler) {
+      throw Error('Cannot collect time profile, time profiler not enabled.');
+    }
+    const durationMillis = parseDuration(prof.duration);
```
aalexand: By the way, did you verify that the parse-duration module supports the text duration format used by the Google API JSON responses?

nolanmar511: Yes -- the responses have formats like '5h5m10s' and that is parsed correctly.
The only possible problem was that this module didn't interpret 'u' as microseconds,  but I didn't see any responses which used 'u'.

PR #17, opened 2017-10-20
clean up profile serialization

nolanmar511:


src/profilers/profile-serializer.ts
```diff
@@ -117,30 +111,36 @@ export function serializeTimeProfile(
...
   function createTimeValueType(): perftools.profiles.ValueType {
-    return new perftools.profiles.ValueType({
-      type: getIndexOrAdd('time', strings),
-      unit: getIndexOrAdd('microseconds', strings)
-    });
+    return new perftools.profiles.ValueType(
+        {type: getIndexOrAdd('time'), unit: getIndexOrAdd('microseconds')});
+  }
+
+  function getIndexOrAdd(str: string): number {
```
aalexand: Remove getIndexOrAdd from utils?

src/profilers/profile-serializer.ts
```diff
@@ -117,30 +111,36 @@ export function serializeTimeProfile(
     if (f !== undefined) {
       return f;
```
aalexand: It's a bit strange that the coverage says this line is not covered. We don't have any two samples that point to a same function?

src/profilers/profile-serializer.ts
```diff
@@ -87,9 +83,7 @@ export function serializeTimeProfile(
           });
           samples.push(sample);
         }
-        for (let child of node.children) {
-          entries.push({node: child, stack: stack.slice(0)});
-        }
+        node.children.map((n) => entries.push({node: n, stack: stack.slice(0)}));
```
kjin: Same comment as above regarding `map`, but you can also take advantage of the spread params operator (that `f(...[a,b,c])` and `f(a,b,c)` are equivalent) here: `entries.push(...node.children.map(n=>({node, stack})))`

src/profilers/profile-serializer.ts
```diff
@@ -69,10 +67,8 @@ export function serializeTimeProfile(
    */
   function serializeNode(root: TimeProfileNode) {
     let entries: Entry[] = [];
```
kjin: Looks like `entries` can be `const`

src/profilers/profile-serializer.ts
nolanmar511: I think this is because locationMap and functionMap use the same key (callUid). So, this code ends up basically being unreachable, because if there is a function, then there is a location, so the location will be returned and getFunction() won't be called.
This isn't problematic now, since there isn't any finer profile resolution than function-level resolution. However, the one-to-one correspondence between functions and locations isn't great, so it's something that I should address in a future PR -- I'll file an issue for it.

nolanmar511:
PTAL

nolanmar511:
I've realized that adding heap profiling and changing the time profile so it doesn't use callUid will require pretty major refactoring.
I'm going to close this -- I think it makes more sense to make some of these changes with other changes.

PR #21, opened 2017-10-23
Update profile.proto

nolanmar511:
Updated profile.proto from github.com/google/pprof.
Updated profile.proto has units field for labels.


third_party/proto/README.google
```diff
@@ -1,5 +1,5 @@
-URL: https://github.com/google/pprof/blob/d174bbe741d0e8039ecf6504b15f708d1ee5c0f2
-Version: d174bbe741d0e8039ecf6504b15f708d1ee5c0f2
+URL: https://github.com/google/pprof/tree/c70cbc2c12c3c07c76269944233837844e7dcb29
+Version: c70cbc2c12c3c07c76269944233837844e7dcb29
```
ofrobots: Do profile.proto or pprof use a versioning scheme? It would be good to be able to use a human readable version here if possible.

nolanmar511: pprof does not have a versioning scheme; I don't believe profile.proto does either.

Issue #13, opened 2017-10-17
Stop using GetCallUid


nolanmar511:
GetCallUid  (used here https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/cpu-profiler.cc#L34) will be deprecated soon. See: https://cs.chromium.org/chromium/src/v8/include/v8-profiler.h?q=v8-profiler&sq=package:chromium&dr&l=203

We should switch to using GetScriptId, GetLineNumber, and GetColumnNumber instead.

PR #25, opened 2017-10-25
feat, fix: update profile serialization for heap profiling

nolanmar511:
Updates profile serialization to work for heap profile.
To allow for code reuse between heap and time profile serialization, also stopped using callUid (addressing #13 and #18).


src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
 import {perftools} from '../profile';
-import {getIndexOrAdd} from '../util';
-import {TimeProfile, TimeProfileNode} from '../v8-types';
-// A stack of function UIDs.
+import {AllocationProfileNode, ProfileNode, TimeProfile, TimeProfileNode} from '../v8-types';
+
+// A stack of function IDs.
 type Stack = Array<number>;
+
+// A function which appends the sample form of entry to samples.
```
aalexand: Can't quite parse "appends the sample form of entry to samples", what does that mean? Could you rephrase?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+    this.strings = [];
+    this.stringsMap = new Map<string, number>();
+  }
 
-  serializeNode(prof.topDownRoot);
+  /**
+   * Returns index of str within the table. Also adds str to string table if
+   * str is not in the table already.
+   *
+   * @param str
```
aalexand: Is there any value in having this empty parameter annotation? Is this the linter requirement or something?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
-    stringTable: strings,
-    // opt drop_frames
-    // opt keep_frames
-    timeNanos: 1000 * 1000 * prof.endTime,                         // Nanos
-    durationNanos: 1000 * 1000 * (prof.endTime - prof.startTime),  // Nanos
+  /**
+   * Returns value type for samples counts (type:samples, units:count), and
+   * adds strings used in this value type to the table.
+   */
+  createSampleValueType(): perftools.profiles.ValueType {
```
aalexand: I am not sure this function and two below should be a part of StringTable class. They only use the public interface of the class and don't seem to clearly belong by functionality, so should be elsewhere imo.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+  while (entries.length > 0) {
+    const entry = entries.pop();
+    if (entry !== undefined) {
+      const node = entry.node;
+      const stack = entry.stack;
+      const location = getLocation(node);
+      stack.unshift(location.id as number);
+      appendToSample(entry, samples);
+      for (let child of node.children) {
+        entries.push({node: child, stack: stack.slice(0)});
```
aalexand: Nit: slice(0) can be just slice()

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+
+  function getLocation(node: ProfileNode): perftools.profiles.Location {
+    const key = {
+      functionKey: {name: node.name, scriptId: node.scriptId},
+      lineNumber: node.lineNumber,
+      columnNumber: node.columnNumber
+    };
+    const keyStr = JSON.stringify(key);
+    let id = locationIdMap.get(keyStr);
+    if (id !== undefined && id <= locations.length) {
```
aalexand: Why do you have the " && id <= locations.length" check? That's an invariant, no?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
-  function getFunction(node: TimeProfileNode): perftools.profiles.Function {
-    const id = node.callUid;
-    let f = functionMap.get(id);
-    if (f !== undefined) {
-      return f;
+  function getFunction(node: ProfileNode): perftools.profiles.Function {
+    const key: FunctionKey = {name: node.name, scriptId: node.scriptId};
+    const keyStr = JSON.stringify(key);
+    let id = functionIdMap.get(keyStr);
+    if (id !== undefined && id <= functions.length) {
```
aalexand: Ditto on the id check.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+  profile.function = functions;
+  profile.stringTable = stringTable.strings;
+
+  function getLocation(node: ProfileNode): perftools.profiles.Location {
+    const key = {
+      functionKey: {name: node.name, scriptId: node.scriptId},
+      lineNumber: node.lineNumber,
+      columnNumber: node.columnNumber
+    };
+    const keyStr = JSON.stringify(key);
```
aalexand: This serialization is a somewhat expensive thing I assume. No cheaper way to achieve the goal?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+      };
 
-  function createTimeValueType(): perftools.profiles.ValueType {
-    return new perftools.profiles.ValueType({
-      type: getIndexOrAdd('time', strings),
-      unit: getIndexOrAdd('microseconds', strings)
-    });
-  }
+  const stringTable = new StringTable();
+  stringTable.getIndexOrAdd('');
```
aalexand: This could be moved to the string table constructor.

src/v8-types.ts
```diff
@@ -24,16 +24,27 @@ export interface TimeProfile {
   startTime: number;
 }
 
-export interface TimeProfileNode {
-  callUid: number;
-  scriptResourceName?: string;
-  functionName?: string;
+export interface ProfileNode {
+  // name could be function name or constructor.
```
aalexand: Nit: I think constructor is a function too so the distinction is subtle here.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,278 @@
...
+export function serializeHeapProfile(
+    prof: AllocationProfileNode, startTimeNanos: number, durationNanos: number,
+    intervalBytes: number): perftools.profiles.IProfile {
+  const appendHeapEntryToSample: AppendEntryToSamples =
+      (entry: AllocationEntry, samples: perftools.profiles.Sample[]) => {
+        if (entry.node.allocations.length > 0) {
+          for (const alloc of entry.node.allocations) {
+            const sample = new perftools.profiles.Sample({
+              locationId: entry.stack,
+              value: [alloc.count, alloc.size * alloc.count]
```
aalexand: size -> sizeBytes

ofrobots:
I would like to see smaller commits addressing each issue atomically. That makes code easier to review and easier to atomically revert (not that this is an urgent need ATM).

Secondly, please start using https://conventionalcommits.org/ for the commit messages.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
-          const sample = new perftools.profiles.Sample({
-            locationId: stack,
-            value: [node.hitCount, node.hitCount * intervalMicros]
-          });
-          samples.push(sample);
-        }
-        for (let child of node.children) {
-          entries.push({node: child, stack: stack.slice(0)});
-        }
+    loc = this.strings.length;
```
ofrobots: `loc = this.strings.push(str) - 1`

src/profilers/profile-serializer.ts
nolanmar511: I don't think so -- I tried using the key directly, but that didn't work.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
+  const functions: Array<perftools.profiles.Function> = [];
+  const locationMap: Map<number, perftools.profiles.Location> = new Map();
+  const functionMap: Map<number, perftools.profiles.Function> = new Map();
+  const functionIdMap = new Map<string, number>();
+  const locationIdMap = new Map<string, number>();
+
+  const entries: Entry[] = root.children.map((n) => ({node: n, stack: []}));
+  while (entries.length > 0) {
+    const entry = entries.pop();
+    if (entry !== undefined) {
```
ofrobots: Tangential: What's the case where `entry` would be `undefined`?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
+ * @param prof - profile to be converted.
+ * @param startTimeNanos - start time of profile, in nanoseconds (POSIX time).
+ * @param durationsNanos - duration of the profile (wall clock time) in
+ * nanoseconds.
+ * @param intervalBytes - bytes allocated between samples.
+ */
+export function serializeHeapProfile(
+    prof: AllocationProfileNode, startTimeNanos: number, durationNanos: number,
+    intervalBytes: number): perftools.profiles.IProfile {
+  const appendHeapEntryToSample: AppendEntryToSamples =
```
ofrobots: The variable name should be pluralized: `appendHeapEntryToSamples`

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
+ * fields.
+ * @param root - root of v8 profile tree describing samples to be appended
+ * to profile.
+ * @param appendToSample - function which converts entry to sample(s)  and
+ * appends these to end of an array of samples.
+ * @param stringTable - string table for the existing profile.
+ */
+function serialize(
+    profile: perftools.profiles.IProfile, root: ProfileNode,
+    appendToSample: AppendEntryToSamples, stringTable: StringTable) {
```
ofrobots: nit: pluralize to `appendToSamples`.

src/v8-types.ts
```diff
@@ -24,16 +24,27 @@ export interface TimeProfile {
...
+  name: string;
+  scriptName: string;
+  scriptId: number;
   lineNumber: number;
+  columnNumber: number;
+  children: Array<ProfileNode>;
+}
+
+export interface TimeProfileNode extends ProfileNode {
+  name: string;
```
ofrobots: Don't this and subsequent properties get inherited from the parent interface?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
+  profile.function = functions;
+  profile.stringTable = stringTable.strings;
+
+  function getLocation(node: ProfileNode): perftools.profiles.Location {
+    const key: LocationKey = {
+      function: {name: node.name, scriptId: node.scriptId},
+      lineNumber: node.lineNumber,
+      columnNumber: node.columnNumber
+    };
+    const keyStr = JSON.stringify(key);
```
ofrobots: This is expensive and overlong. How about 

```
const keyStr = `${scriptId}:${lineNumber}:${columnNumber}:${name}`;
```

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
   }
 
-  function getFunction(node: TimeProfileNode): perftools.profiles.Function {
-    const id = node.callUid;
-    let f = functionMap.get(id);
-    if (f !== undefined) {
-      return f;
+  function getFunction(node: ProfileNode): perftools.profiles.Function {
+    const key: FunctionKey = {name: node.name, scriptId: node.scriptId};
+    const keyStr = JSON.stringify(key);
```
ofrobots: likewise.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,279 @@
...
+/**
+ * Converts v8 time profile into into a profile proto.
+ * (https://github.com/google/pprof/blob/master/proto/profile.proto)
+ *
+ * @param prof - profile to be converted.
+ * @param intervalMicros - average time (microseconds) between samples.
+ */
+export function serializeTimeProfile(
+    prof: TimeProfile, intervalMicros: number): perftools.profiles.IProfile {
+  const appendTimeEntryToSample: AppendEntryToSamples =
```
ofrobots: pluralize

src/v8-types.ts
```diff
@@ -24,16 +24,27 @@ export interface TimeProfile {
...
+  scriptName: string;
+  scriptId: number;
+  lineNumber: number;
+  columnNumber: number;
   hitCount: number;
   children: Array<TimeProfileNode>;
 }
 
-export interface AllocationProfileNode {
+export interface AllocationProfileNode extends ProfileNode {
```
ofrobots: likewise.

src/profilers/profile-serializer.ts
nolanmar511: entry should never be undefined, but if I remove that if statement, then I get the error 
"[ts] Object is possibly 'undefined'.
const entry: Entry | undefined"

Is there a more standard way to deal with that?

nolanmar511:
PTAL

src/profilers/profile-serializer.ts
ofrobots: Given that you're checking for length as the loop condition, entry can never be undefined, but the compiler doesn't know that constraint. You can either use a cast or use (type guards)[https://www.typescriptlang.org/docs/handbook/advanced-types.html].

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,260 @@
...
 import {perftools} from '../profile';
-import {getIndexOrAdd} from '../util';
-import {TimeProfile, TimeProfileNode} from '../v8-types';
-// A stack of function UIDs.
+import {AllocationProfileNode, ProfileNode, TimeProfile, TimeProfileNode} from '../v8-types';
+
+// A stack of function IDs.
 type Stack = Array<number>;
+
+// A function which converts entry into one or more samples, then
```
ofrobots: nit: prefer jsdoc style comments here and for the types below.

nolanmar511:
@aalexand - any comments?

aalexand:
@nolanmar511 I had some comments, I will go over the source now to see if they were addressed. It would be an easier job for me if you'd respond "Done" to things that are addressed so that I know they have been addressed. I tend to skip verifying minor things in that case just trusting the author.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,270 @@
...
+   * Returns index of str within the table. Also adds str to string table if
+   * str is not in the table already.
    */
-  function serializeNode(root: TimeProfileNode) {
-    let entries: Entry[] = [];
-    // don't include root node in serialized profile, start with it's children.
-    for (let child of root.children) {
-      entries.push({node: child, stack: []});
+  getIndexOrAdd(str: string): number {
+    let loc = this.stringsMap.get(str);
```
aalexand: Nit: loc -> idx since that's how you call that in the function doc.

src/profilers/profile-serializer.ts
aalexand: @ofrobots Would "const entry = entries.pop()!;" work?

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,270 @@
...
+  const locationIdMap = new Map<string, number>();
+
+  const entries: Entry[] = root.children.map((n) => ({node: n, stack: []}));
+  while (entries.length > 0) {
+    const entry = entries.pop();
+    if (entry !== undefined) {
+      const node = entry.node;
+      const stack = entry.stack;
+      const location = getLocation(node);
+      stack.unshift(location.id as number);
```
aalexand: Would "stack.unshift(location.id!);" work? As far as I understand that discards the "undefined | null" portion of the type which is more type-safe than a cast.

src/profilers/profile-serializer.ts
```diff
@@ -15,132 +15,270 @@
...
+  function getLocation(node: ProfileNode): perftools.profiles.Location {
+    const keyStr =
+        `${node.scriptId}:${node.lineNumber}:${node.columnNumber}:${node.name}`;
+    let id = locationIdMap.get(keyStr);
+    if (id !== undefined) {
+      // id is index+1, since 0 is not valid id.
+      return locations[id - 1];
     }
-    location = new perftools.profiles.Location({id: id, line: [getLine(node)]});
+    id = locations.length + 1;
```
aalexand: Nit: `id = locations.push(location) + 1;` and remove the corresponding push line below?

src/profilers/profile-serializer.ts
nolanmar511: No, I get the following error if I remove the cast:
[ts] Argument of type 'number | Long' is not assignable to parameter of type 'number'.
  Type 'Long' is not assignable to type 'number'

src/profilers/profile-serializer.ts
nolanmar511: I'm not sure if that would be clear, since `id` is a field in `location`.
I could change the id after creating `location` but I think that would be unclear.

src/profilers/profile-serializer.ts
nolanmar511: Done

src/profilers/profile-serializer.ts
aalexand: @nolanmar511 Did you try what I suggested?

nolanmar511: Just did -- that works.

ofrobots:
Whoa, I didn't know about the non-null assertion operator. That's a much more readable cast, which is great.

In some cases (although probably not here), it is possible use a type-guard to avoid a cast, and that would still be preferable if possible.

nolanmar511:
Could this be merged in now?

aalexand:
I don't have "Merge pull request" button enabled. There is text next to the button that says "You're not authorized to merge this pull request." @ofrobots Can you merge?

ofrobots:
The protected branch configuration was enabled recently on all our repos – that was the cause. Created a new team @GoogleCloudPlatform/node-profiler-team, add you both to it, and gave it the ability to land things on protected branches. Let me know if the merge button is still disabled.

aalexand:
@ofrobots I would also disable the "Rebase and merge" option on the repo. "Squash and merge" is the only sane choice and should be the only one allowed imo.

ofrobots:
Done.

PR #29, opened 2017-10-27
chore: reorganize source code

ofrobots:
Reorganize the source code so that we no longer change the path 
relative to the root directory as part of the compilation process. 

* The output from proto file compilation is now checked into the repo.
  The proto compile step does not need to be run on each compile/test.
  Instead we can regenerate the output files whenever the proto is
  updated. For this to work, we needed to ensure that the relative
  path to the generated files remains the same between compile time and
  runtime – these files are needed at both times.
* TypeScript wants all source files to be in the `rootDir` which
  necessitated the move of third_party/types. Hopefully we can get rid
  of this file completely soon.
* Drop types/gcp-metadata-types.ts as gcp-metadata now ships with
  types.
* Drop third_party from package.json/files. Not sure why it needed to
  shipped.

Open questions:
* I don't like the name for the top-level `dev` directory. Suggestions welcome.

~~If this works out I propose we use this as the standard layout for gts-based projects.~~

R=@nolanmar511 @GoogleCloudPlatform/node-team


nolanmar511:
My only concern is that types got moved out of third party.

ofrobots:
@nolanmar511 valid point. I've addressed that be adding back the `third_party` prefix to `types`. TypeScript wants all of this source code to be be in the `rootDir` so it has to move into `dev`.

Aside, we should add this type definition file to the upstream module so that we can drop it from third_party here and in the trace agent. That might be a bit faster than migrating upstream to typescript.

aalexand:
@ofrobots Sorry, I didn't 100% get the necessity of the "dev" subdirectory and it looks somewhat clumsy so I thought it needs to be challenged a bit - can you elaborate why it's needed?

ofrobots:
@aalexand that's a fair question. `dev` is clunky indeed. This was discussed in chat, but you're right that the details should be here as well.

* Basically, the way the code is laid out, the path to a file changes depth between compile time and runtime. For example, `src/foo.ts`, once compiled, becomes `build/src/foo.js`.
* This means that if the file foo needs to refer to an asset or fixture, the relative path is different between the source files and the output files. So if `src/foo.js` needed to refer to `fixtures/bar` the relative path needs to be `../../fixtures/bar` because the runtime path is what actually matters in most cases. In practice this is the case already as we need to load the `package.json` and must use the path relative to the output file rather the source file. This is annoying but, not a deal breaker.
* Now if you have an asset that you need *both* at compile time *and* at runtime, this breaks down. There isn't a single relative path that can refer to the correct file at both times. More concretely, the proto-generated js and d.ts files are needed at both times.
* Symlinks don't help here, as the runtime relative path must be the canonical one, and in order to achieve that, we will need to install a symlink in the parent of the root directory (not acceptable).
* We are presently working around this by generating a copy of the needed file in both potential locations. This is very clunky. 
* At the moment there are only two files that fall into this category, but we need to bring in new proto files which make this solution quite terrible.

Creating a source root directory (`dev` in this PR) works around this problem by making sure that the relative path to fixtures is constant between runtime and compile time. While still clunky, I think it is preferable to the differing paths to fixtures.

aalexand:
@ofrobots Would it be possible to generate build/src/foo.js as src/foo.js instead?

ofrobots:
@aalexand it is possible, but we would really like to have the output fix sequestered into a separate directory. This ensures that we can trust `npm publish` and `npm run clean` to only touch the necessary files. Cluttering the source directory with output files is more clunky IMO.

aalexand:
@ofrobots I see. Could the ts compiler run in src directory and generate the output files as build/foo.js and so on?

If no, how about renaming "dev" to "typescript" or "ts"?

ofrobots:
@aalexand Then you end up with `src/` and `test/` files intermingling (and potentially colliding) in `build/`. Further more, we would need to erase the test files from `build/` before publishing.

The suggestion to rename `dev` to `ts` was good. Done.

aalexand:
LGTM

PR #31, opened 2017-10-28
Import googleapis

ofrobots:
**DO NOT LAND USING GITHUB UI**

This imports googleapis project as a [git-subtree](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree). We need `profiler.proto` and other proto files from that project here.

git-subtree makes it easy to maintain remote third_party content. README.google contains instructions on how to keep the remote up to date.


aalexand:
LGTM overall, curious why git-subtree and not a submodule? (haven't used the former, used the latter)

ofrobots:
Git submodules have a lot of gotchas (see [link](https://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/)), but the biggest one for me are the fact that 1) a freshly cloned repo is not  a complete checkout, increasing the barrier to new contributors to the project, and 2) even developers have to remember to `git submodule update` upon checkout of the main repo.

A subtree, OTOH, is a sub directory of files originating from a  remote repo checked into the main repo. There is only one git repo. They are an excellent match for the "third_party" concept as we use it.

PR #33, opened 2017-10-31
fix: remove build/ from nyc exclude

ofrobots:
Coverage reporting regressed by #29 which added build/ to nyc.exclude
configuration.


ofrobots:
TBR= @nolanmar511

PR #34, opened 2017-10-31
feat: add heap profile collection

nolanmar511:


ts/src/profilers/heap-profiler.ts
```diff
@@ -1,7 +1,58 @@
...
+import {perftools} from '../../../proto/profile';
+import {serializeHeapProfile} from './profile-serializer';
+
+const profiler = require('bindings')('sampling_heap_profiler');
+
 export class HeapProfiler {
-  constructor() {
-    throw new Error('Heap Profiler is unimplemented');
+  private enabled: boolean;
+  /**
```
aalexand: Nit: add an empty line before this one.

ts/src/profilers/heap-profiler.ts
```diff
@@ -1,7 +1,58 @@
...
+   */
+  constructor(private intervalBytes: number, private stackDepth: number) {
+    this.enable();
+  }
+
+  /**
+   * Collects a heap profile when heapProfiler is enabled. Otherwise throws
+   * an error.
+   *
+   * @param durationMillis - recorded duration of the profile.
```
aalexand: Heap profiles do not have duration, they are instantaneous.

ts/src/profilers/heap-profiler.ts
```diff
@@ -1,7 +1,58 @@
...
+   */
+  profile(durationMillis: number): perftools.profiles.IProfile {
+    if (!this.enabled) {
+      throw new Error('Heap profiler not enabled.');
```
aalexand: Nit: "... is not enabled." to make the message a proper sentence.

ts/src/profilers/heap-profiler.ts
```diff
@@ -1,7 +1,58 @@
...
+    }
+    const result = profiler.getAllocationProfile();
+    const startTimeNanos = Date.now() * 1000 * 1000;
+    return serializeHeapProfile(
+        result, startTimeNanos, durationMillis * 1000 * 1000,
```
aalexand: See above on duration.

nolanmar511: Done.

ts/src/profilers/heap-profiler.ts
nolanmar511: Done.

ts/src/profilers/heap-profiler.ts
nolanmar511: Done.

ts/src/profilers/heap-profiler.ts
nolanmar511: Done.

ts/src/profilers/heap-profiler.ts
```diff
@@ -1,7 +1,55 @@
...
+    this.enable();
+  }
+
+  /**
+   * Collects a heap profile when heapProfiler is enabled. Otherwise throws
+   * an error.
+   */
+  profile(): perftools.profiles.IProfile {
+    if (!this.enabled) {
+      throw new Error('Heap profiler is not enabled.');
```
aalexand: Do we catch such errors at the top level and make sure they just get logged into the profiler log and do not kill the customer's program?

Note that it's possible that customer starts their program with heap enabled on one machine and with heap disabled on another machine and the current behavior of the server is that the set of profile types is the same among all clients of the same deployment so the one with heap profiling disabled could be asked by the server to collect the heap data and the agent should ignore that logging a warning.

nolanmar511: Yes -- these errors are caught at the top level and logged.

aalexand:
@ofrobots Any comments on the code?

ts/test/test-heap-profiler.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    after(() => {
+      v8HeapProfiler.startSamplingHeapProfiler.restore();
+      v8HeapProfiler.stopSamplingHeapProfiler.restore();
+      v8HeapProfiler.getAllocationProfile.restore();
+      (Date.now as any).restore();
+    });
+
+    it('should return a profile equal to the expected profile', async () => {
+      const durationMillis = 10 * 1000;
```
ofrobots: Unused?

ts/test/test-heap-profiler.ts
```diff
@@ -0,0 +1,67 @@
...
+      const intervalBytes = 1024 * 512;
+      const stackDepth = 32;
+      let profiler = new HeapProfiler(intervalBytes, stackDepth);
+      let profile = profiler.profile();
+      assert.deepEqual(heapProfile, profile);
+    });
+
+    it('should throw error when disabled', async () => {
+      const durationMillis = 10 * 1000;
```
ofrobots: unused? If gts didn't catch this, can you open a bug against gts?

ts/test/test-heap-profiler.ts
```diff
@@ -0,0 +1,67 @@
...
+      const intervalBytes = 1024 * 512;
+      const stackDepth = 32;
+      const profiler = new HeapProfiler(intervalBytes, stackDepth);
+      profiler.disable();
+      try {
```
ofrobots: You can use [`assert.throws`](https://nodejs.org/dist/latest-v8.x/docs/api/assert.html#assert_assert_throws_block_error_message) for asserting that a snippet of code throws.

nolanmar511: Does `assert.throws` work for asynchronous things?

I tried adding the following and the test failed.
```
assert.throws(
        async () => {
          throw new Error('ERROR');
        },
        Error
      );
```

But this passed:
```
assert.throws(
         () => {
          throw new Error('ERROR');
        },
        Error
      );
```

nolanmar511: Never mind -- I forget that profiling isn't asynchronous for heap-profiler.

PR #35, opened 2017-10-31
fix: ignore build/test in test coverage

nolanmar511:
Including tests in code coverage report seems to be inflating the overall code coverage.
I feel that this makes the code coverage report misleading, so tests should not be included in code coverage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   ],
   "nyc": {
     "exclude": [
-      "proto"
+      "proto",
+      "build/test"
     ]
   }
 }


ofrobots:
Is there a strong reason to exclude the test directory from coverage reports? I can see value in knowing if the tests code is actually getting exercises as well. However, if this is causing grief, let's be document the reason for excluding the test directory in this PR.

aalexand:
I sort of like seeing that the tests get executed, but I think it's better to have the coverage number represent the cover over the code excluding the tests.

nolanmar511:
Especially considering that we currently do have checks on the code coverage in our PRs, I think that it's misleading to include the tests in the code coverage, especially since this seems to inflate the overall code coverage.  
It seems like now if I copy-and-pasted the same test 100 times and that test had 100% coverage, then I could drive up the overall test coverage (by increasing the coverage just for tests, not for the actual code). That seems wrong to me.

PR #37, opened 2017-11-01
fix: add support for TypeScript 2.6

ofrobots:
TypeScript 2.6 makes type checking for functions stricter [1]. Instead
of using bivariant types for functions parameters, TypeScript now uses
contravaiance.

This has been breaking our CI with Node 6 ([example](https://travis-ci.org/GoogleCloudPlatform/cloud-profiler-nodejs/jobs/295479690)) as it doesn't use the locked down version of typescript as specified by our `package-lock.json`. Let's bite the bullet and upgrade to TypeScript 2.6.

[1] https://blogs.msdn.microsoft.com/typescript/2017/10/31/announcing-typescript-2-6/

/cc @matthewloring in case he has ideas on how to avoid the cast to `Array<T>` for the `children` array.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,13 +38,13 @@
     "@protobufjs/base64": {
       "version": "1.1.2",
       "resolved": "https://registry.npmjs.org/@protobufjs/base64/-/base64-1.1.2.tgz",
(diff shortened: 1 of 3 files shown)


matthewloring:
Short answer: https://github.com/matthewloring/cloud-profiler-nodejs/commit/4c1aa175e747012683728267cfc95bbf7f61ab27

Long answer: https://www.cs.cmu.edu/~aldrich/courses/819/f-bounded.pdf

PR #39, opened 2017-11-01
fix: should use '(anonymous)' as function name when no name specified

nolanmar511:
This is a regression from #34.
If no function name is specified, the function name should be recorded as  '(anonymous)' .

### DIFF
--- a/ts/src/profilers/profile-serializer.ts
+++ b/ts/src/profilers/profile-serializer.ts
@@ -140,7 +140,7 @@ function serialize<T extends ProfileNode>(
     }
     id = functions.length + 1;
     functionIdMap.set(keyStr, id);
-    let nameId = stringTable.getIndexOrAdd(node.name);
+    let nameId = stringTable.getIndexOrAdd(node.name || '(anonymous)');
     let f = new perftools.profiles.Function({
       id: id,
       name: nameId,
--- a/ts/src/v8-types.ts
+++ b/ts/src/v8-types.ts
@@ -26,7 +26,7 @@ export interface TimeProfile {
 
 export interface ProfileNode {
   // name is the function name.
-  name: string;
+  name?: string;
   scriptName: string;
   scriptId: number;
   lineNumber: number;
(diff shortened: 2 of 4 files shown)


nolanmar511:
PTAL

ts/src/profilers/profile-serializer.ts
```diff
@@ -140,7 +140,7 @@ function serialize<T extends ProfileNode>(
     }
     id = functions.length + 1;
     functionIdMap.set(keyStr, id);
-    let nameId = stringTable.getIndexOrAdd(node.name);
+    let nameId = stringTable.getIndexOrAdd(node.name || '(anonymous)');
```
aalexand: The only thing that slightly worries me is whether empty string always means anonymous function. Are there conditions when the runtime actually failed to provide the function name for some other reason? Should we rename the function to (anonymous) only if the script name is present, or that's overkill?

ofrobots: /cc @hashseed: this is probably related to function name inference done by V8. What are the cases when we get an empty function name vs. `(anonymous)`?

hashseed: I don't think the function name inferrer ever returns "(anonymous)". It would be empty.

aalexand: @hashseed Is there a way to tell "good empty" (i.e. anonymous function) from "bad empty" (i.e. something failed while the function name was being inferred)?

ofrobots: Looking at the V8 source code, I don't see the string `(anonymous)` anywhere. I am guessing that V8 returns empty string in anonymous case.

Looking at the DevTools source, it seems they use the [same logic](https://cs.chromium.org/chromium/src/out/Debug/gen/devtools/inspector.js?type=cs&q=%5C(anonymous%5C)&l=2997) as being proposed in this PR: infer `(anonymous)` when the name is empty.

PR #40, opened 2017-11-01
fix: change sample type for heap profile

nolanmar511:
Sample type for first value in heap profile was incorrect, so the profile could not be uploaded.
This change fixes that.

### DIFF
--- a/ts/src/profilers/profile-serializer.ts
+++ b/ts/src/profilers/profile-serializer.ts
@@ -171,7 +171,7 @@ function serialize(
  * Returns value type for samples counts (type:samples, units:count), and
  * adds strings used in this value type to the table.
  */
-function createSampleValueType(table: StringTable):
+function createTimeCountValueType(table: StringTable):
     perftools.profiles.ValueType {
   return new perftools.profiles.ValueType({
     type: table.getIndexOrAdd('samples'),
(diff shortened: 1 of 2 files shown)


ts/src/profilers/profile-serializer.ts
```diff
@@ -222,7 +234,7 @@ export function serializeTimeProfile(
       };
 
   const stringTable = new StringTable();
-  const sampleValueType = createSampleValueType(stringTable);
+  const sampleValueType = createTimeCountValueType(stringTable);
```
aalexand: "time count" sounds weird. Perhaps "sample count"?

nolanmar511: done

ts/src/profilers/profile-serializer.ts
```diff
@@ -191,6 +191,18 @@ function createTimeValueType(table: StringTable): perftools.profiles.ValueType {
 }
 
 /**
+ * Returns value type for object counts (type:objects, units:count), and
+ * adds strings used in this value type to the table.
+ */
+function createAllocationCountValueType(table: StringTable):
```
aalexand: I would name it createObjectCountValueType since that's what it is according to the doc and the source code.

PR #52, opened 2017-11-13
change location for compiled typescript

nolanmar511:
node-gyp will delete the 'build' folder on install.
This PR moves the compiled typescript to a different folder, so the compiled typescript does not also get deleted.


package.json
```diff
@@ -3,11 +3,11 @@
   "version": "0.1.0",
   "description": "Adds support for Google Cloud Profiler to node.js applications",
   "repository": "GoogleCloudPlatform/cloud-profiler-nodejs",
-  "main": "build/src/index.js",
-  "types": "build/src/index.d.ts",
+  "main": "compiled-ts/src/index.js",
```
ofrobots: How about `out/`?

PR #53, opened 2017-11-13
add out/src to files

nolanmar511:
After #52, npm package did not include src files and code coverage of tests was added back to code coverage.
This change fixes that.


nolanmar511:
PTAL

PR #57, opened 2017-11-16
don't send instance and zone if values not specified

### LINKED ISSUE #55: Don't send instance or zone if there is no value.
Don't send instance or zone if there is no value.

nolanmar511:
Fixes #55

### DIFF
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -87,8 +87,8 @@ export interface ProfilerConfig extends AuthenticationConfig {
   projectId?: string;
   logLevel: number;
   serviceContext: {service: string; version?: string;};
-  instance: string;
-  zone: string;
+  instance?: string;
+  zone?: string;
   disableTime: boolean;
   disableHeap: boolean;
   timeIntervalMicros: number;
@@ -103,8 +103,6 @@ export const defaultConfig = {
   serviceContext: {},
   disableHeap: false,
   disableTime: false,
-  instance: '',
-  zone: '',
   timeIntervalMicros: 1000,
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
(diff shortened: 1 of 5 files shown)


ts/src/config.ts
```diff
@@ -87,8 +87,8 @@ export interface ProfilerConfig extends AuthenticationConfig {
   projectId?: string;
   logLevel: number;
   serviceContext: {service: string; version?: string;};
-  instance: string;
-  zone: string;
+  instance?: string;
```
aalexand: I would keep the type and skip sending if the string is empty - what's wrong with that? We don't want to send empty strings either and I am not sure the current code does not?

nolanmar511: Right now, the code will skip sending these if empty.
I've added a test for that now.

ts/src/index.ts
```diff
@@ -84,11 +84,11 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
                 (err: Error) => {
                     // ignore errors, which will occur when not on GCE.
                 }) ||
-        ['', ''];
-    if (!mergedConfig.zone) {
+        [undefined, undefined];
```
ofrobots: `[]` would be adequate.

nolanmar511: I tried that, but got the following error:

ts/src/index.ts(78,12): error TS2525: Initializer provides no value for this binding element and the binding element has no default value.

78     const [instance, zone] =

nolanmar511:
PTAL

aalexand:
@nolanmar511 Merge this?

aalexand:
The description should say that this fixes https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/issues/55 ?

PR #69, opened 2017-11-20
updated logged messages

nolanmar511:


ts/src/profiler.ts
```diff
@@ -282,10 +281,11 @@ export class Profiler extends common.ServiceObject {
       }
       throw new Error(message.toString());
     }
-    if (!isRequestProfile(body)) {
-      throw new Error(`Profile not valid: ${body}.`);
+    if (!isRequestProfile(prof)) {
+      throw new Error(`Profile not valid: ${prof}.`);
     }
-    return body;
+    this.logger.debug(`Successfully created profile ${prof.profileType}`);
```
aalexand: Can you add a debug message (if there is no one, and I think there isn't) before the create profiler call is started?

ts/src/profiler.ts
```diff
@@ -297,6 +298,7 @@ export class Profiler extends common.ServiceObject {
    */
   async profileAndUpload(prof: RequestProfile): Promise<void> {
     prof = await this.profile(prof);
+    this.logger.debug(`Successfully collected profile ${prof.profileType}`);
```
aalexand: Nit: this and some other log messages do not have the full stop, while others have. Please be consistent. I think it is common to omit the stop when the whole message is a single sentence and use the full stops when it's several sentences.

ts/src/profiler.ts
```diff
@@ -228,14 +228,13 @@ export class Profiler extends common.ServiceObject {
...
-      this.logger.error(
-          `Error requesting profile type to be collected: ${err}.`);
+      this.logger.error(`Failed to create profile: ${err}.`);
       return this.config.backoffMillis;
     }
     try {
       await this.profileAndUpload(prof);
     } catch (err) {
-      this.logger.error(`Error collecting and uploading profile: ${err}.`);
+      this.logger.error(`Failed to collect and upload profile: ${err}.`);
```
nolanmar511: @aalexand -- should I do a full-stop here?
Some errors don't end with a full stop and some do. Should I add a full stop only if the error doesn't have one? Or put the error in quotes?

aalexand: I would drop the full stop here.

Also, if the error message starts with a capital letter, the text will read weird here with the capital letter after the colon: `Failed to do foo: Failed to connect to server bar.`. This is why Go has the convention that the error messages start with a lowercase, so that the chaining them in the log messages looks good. I don't know whether there is a Node.js style for this, but if there is we should follow it.

nolanmar511: I think I started with lower-case error messages, and changed because Node.js convention is to capitalize errors. Errors I've seen generated by other code are capitalized.

PR #73, opened 2017-11-21
chore: add test for starting and running agent.

nolanmar511:


aalexand:
Is there a reason you updated against the master using the merge and not the rebase? Seeing these merge commits in a pull request is super weird.

nolanmar511:
Fixed commit history.

ofrobots:
An end to end test should talk to the real profiler service rather than a mocked service. Is there a reason why this wasn't done?

aalexand:
@ofrobots That non-hermetic test should be a separate one.

aalexand:
Anything holding this?

PR #74, opened 2017-11-21
fix: make duration in RequestProfile optional, so Heap Profiles do not have to have durations

nolanmar511:
Right now, duration is a required field of RequestProfile. This does not make sense because heap profiles do not need to have a duration.

Prior to this change, when attempting to collect heap profiles, the following error would occur:
```
Error collecting and uploading profile: ApiError: Request contains an invalid argument
```

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -66,7 +66,7 @@ export interface Deployment {
 export interface RequestProfile {
   name: string;
   profileType?: string;
-  duration: string;
+  duration?: string;
   profileBytes?: string;
   deployment?: Deployment;
   labels?: {instance?: string};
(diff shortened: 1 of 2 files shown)


aalexand:
For fixes, the description should carry the details of what is being fixed or link to the issue being fixed. From just seeing the title of "fix: make duration optional" I can't see why this is done.

ts/src/profiler.ts
```diff
@@ -80,12 +80,7 @@ function isDeployment(deployment: any): deployment is Deployment {
   return (deployment.projectId === undefined ||
```
aalexand: Coverage says this is not covered by tests?

ts/test/test-profiler.ts
```diff
@@ -383,6 +380,25 @@ describe('Profiler', () => {
          assert.ok(
              requestProfileMock.isDone(), 'expected call to create profile');
        });
+    it('should receive heap profile response without error.', async () => {
```
aalexand: Drop trailing full stop in the test description here and in all other places where you have it. See https://jasmine.github.io/2.1/introduction.html, these do not have the full stops.

nolanmar511: Fixed, at least within modified and new tests here.

ts/src/profiler.ts
nolanmar511: Adjusted tests and fixed this.

nolanmar511:
PTAL

ofrobots:
Note that the description should explain why the change is being made. You didn't need to make the abstract (the first line) verbose.

The abstract is what is changing. The description is why it is changing.

Issue #44, opened 2017-11-02
The backoff on profile creation errors needs to be exponential and randomized


aalexand:
Use the same factor, randomization and capping as the Go and Java agents.

nolanmar511:
Fixed by #75

PR #75, opened 2017-11-21
feat: add exponential and server-mediated backoff on retries

### LINKED ISSUE #43: Handle server-instructed backoff duration

nolanmar511:
Fixes #43 and #44.
See #56 for previous discussion on this topic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "nan": "^2.8.0",
     "parse-duration": "^0.1.1",
     "pify": "^3.0.0",
+    "pretty-ms": "^3.1.0",
     "protobufjs": "~6.8.0",
     "request": "^2.83.0",
     "retry-request": "^3.0.1"
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google-cloud/common": "^0.15.0",
-    "@types/pretty-ms": "^3.0.0",
     "bindings": "^1.2.1",
     "delay": "^2.0.0",
     "extend": "^3.0.1",
@@ -47,6 +46,7 @@
     "@types/nock": "^8.2.1",
     "@types/node": "^8.0.30",
     "@types/pify": "^3.0.0",
+    "@types/pretty-ms": "^3.0.0",
     "@types/request": "^2.0.7",
     "@types/sinon": "^4.0.0",
     "clang-format": "^1.0.33",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google-cloud/common": "^0.15.0",
+    "@types/pretty-ms": "^3.0.0",
     "bindings": "^1.2.1",
     "delay": "^2.0.0",
     "extend": "^3.0.1",
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -99,8 +99,9 @@ export interface Config extends AuthenticationConfig {
   // Server-specified backoffs will be capped at serverBackoffCapMillis.
   // The backoff is capped here because setTimeout (which is used to control
   // when next profile is collected) will run immediately if the backoff is
-  // to large.
-  serverBackoffCapMillis?: number;
+  // too large.
+  // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
+  serverBackoffCapMillis?: 2147483646;
 }
 
 // Interface for an initialized config.
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -79,25 +79,28 @@ export interface Config extends AuthenticationConfig {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
+  // On each consecutive error in profile creation, the backoff envelope will
+  // increase by this factor. The backoff will be a random value selected
+  // from a uniform distribution between 0 and the backoff envelope.
+  backoffMultiplier?: number;
+
   // On first error during profile creation, if the backoff is not specified
-  // by the server response, then profiler will between 0 and
+  // by the server response, then profiler will wait between 0 and
   // initialBackoffMillis before asking the server to create a profile again.
-  // After a successful profile creation, the backoff will be reset to
+  // After a successful profile creation, the backoff envelope will be reset to
   // initialExpBackoffMillis.
   initialBackoffMillis?: number;
 
   // If the backoff is not specified by the server response, then profiler will
-  // wait at most maxBackoffMillis before asking server to create a profile
+  // wait at most expBackoffMillisCap before asking server to create a profile
   // again.
-  maxBackoffMillis?: number;
+  expBackoffMillisCap?: number;
 
-  // On each consecutive error in profile creation, the maximum backoff will
-  // increase by this factor. The backoff will be random value selected
-  // from a uniform distribution between 0 and the maximum backoff.
-  backoffMultiplier?: number;
-
-  // Server-specified backoffs will be capped at backoffLimitMillis.
-  backoffLimitMillis?: number;
+  // Server-specified backoffs will be capped at serverBackoffMillisCap.
+  // The backoff is capped here because setTimeout (which is used to controll
+  // when next profile is collected) will run immediately if the backoff is
+  // to large.
+  serverBackoffMillisCap?: number;
 }
 
 // Interface for an initialized config.
@@ -113,9 +116,9 @@ export interface ProfilerConfig extends AuthenticationConfig {
   heapIntervalBytes: number;
   heapMaxStackDepth: number;
   initialBackoffMillis: number;
-  maxBackoffMillis: number;
+  expBackoffMillisCap: number;
   backoffMultiplier: number;
-  backoffLimitMillis: number;
+  serverBackoffMillisCap: number;
 }
 
 // Default values for configuration for a profiler.
@@ -128,7 +131,7 @@ export const defaultConfig = {
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
   initialBackoffMillis: 1000,
-  maxBackoffMillis: parseDuration('1h'),
+  expBackoffMillisCap: parseDuration('1h'),
   backoffMultiplier: 1.3,
-  backoffLimitMillis: parseDuration('7d'),  // 7 days
+  serverBackoffMillisCap: parseDuration('7d'),  // 7 days
 };
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -97,7 +97,7 @@ export interface Config extends AuthenticationConfig {
   expBackoffMillisCap?: number;
 
   // Server-specified backoffs will be capped at serverBackoffMillisCap.
-  // The backoff is capped here because setTimeout (which is used to controll
+  // The backoff is capped here because setTimeout (which is used to control
   // when next profile is collected) will run immediately if the backoff is
   // to large.
   serverBackoffMillisCap?: number;
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -101,7 +101,7 @@ export interface Config extends AuthenticationConfig {
   // when next profile is collected) will run immediately if the backoff is
   // too large.
   // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
-  serverBackoffCapMillis?: 2147483646;
+  serverBackoffCapMillis?: number;
 }
 
 // Interface for an initialized config.
@@ -134,5 +134,5 @@ export const defaultConfig = {
   initialBackoffMillis: 1000,
   backoffCapMillis: parseDuration('1h'),
   backoffMultiplier: 1.3,
-  serverBackoffCapMillis: parseDuration('7d')
+  serverBackoffCapMillis: 2147483647
 };
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -134,5 +134,9 @@ export const defaultConfig = {
   initialBackoffMillis: 1000,
   backoffCapMillis: parseDuration('1h'),
   backoffMultiplier: 1.3,
+
+  // This is the largest duration for setTimeout which does not cause it to
+  // run immediately.
+  // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
   serverBackoffCapMillis: 2147483647
 };
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -88,19 +88,19 @@ export interface Config extends AuthenticationConfig {
   // by the server response, then profiler will wait between 0 and
   // initialBackoffMillis before asking the server to create a profile again.
   // After a successful profile creation, the backoff envelope will be reset to
-  // initialExpBackoffMillis.
+  // initialBackoffMillis.
   initialBackoffMillis?: number;
 
   // If the backoff is not specified by the server response, then profiler will
-  // wait at most expBackoffMillisCap before asking server to create a profile
+  // wait at most backoffCapMillis before asking server to create a profile
   // again.
-  expBackoffMillisCap?: number;
+  backoffCapMillis?: number;
 
-  // Server-specified backoffs will be capped at serverBackoffMillisCap.
+  // Server-specified backoffs will be capped at serverBackoffCapMillis.
   // The backoff is capped here because setTimeout (which is used to control
   // when next profile is collected) will run immediately if the backoff is
   // to large.
-  serverBackoffMillisCap?: number;
+  serverBackoffCapMillis?: number;
 }
 
 // Interface for an initialized config.
@@ -116,9 +116,9 @@ export interface ProfilerConfig extends AuthenticationConfig {
   heapIntervalBytes: number;
   heapMaxStackDepth: number;
   initialBackoffMillis: number;
-  expBackoffMillisCap: number;
+  backoffCapMillis: number;
   backoffMultiplier: number;
-  serverBackoffMillisCap: number;
+  serverBackoffCapMillis: number;
 }
 
 // Default values for configuration for a profiler.
@@ -131,7 +131,7 @@ export const defaultConfig = {
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
   initialBackoffMillis: 1000,
-  expBackoffMillisCap: parseDuration('1h'),
+  backoffCapMillis: parseDuration('1h'),
   backoffMultiplier: 1.3,
-  serverBackoffMillisCap: parseDuration('7d'),  // 7 days
+  serverBackoffCapMillis: parseDuration('7d')
 };
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -16,6 +16,7 @@
 
 import {AuthenticationConfig, Common, ServiceConfig} from '../third_party/types/common-types';
 
+const parseDuration: (str: string) => number = require('parse-duration');
 const common: Common = require('@google-cloud/common');
 const extend = require('extend');
 
@@ -78,9 +79,25 @@ export interface Config extends AuthenticationConfig {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
-  // Time to wait before trying to create a profile again if profile creation
-  // fails.
-  backoffMillis?: number;
+  // On first error during profile creation, if the backoff is not specified
+  // by the server response, then profiler will between 0 and
+  // initialBackoffMillis before asking the server to create a profile again.
+  // After a successful profile creation, the backoff will be reset to
+  // initialExpBackoffMillis.
+  initialBackoffMillis?: number;
+
+  // If the backoff is not specified by the server response, then profiler will
+  // wait at most maxBackoffMillis before asking server to create a profile
+  // again.
+  maxBackoffMillis?: number;
+
+  // On each consecutive error in profile creation, the maximum backoff will
+  // increase by this factor. The backoff will be random value selected
+  // from a uniform distribution between 0 and the maximum backoff.
+  backoffMultiplier?: number;
+
+  // Server-specified backoffs will be capped at backoffLimitMillis.
+  backoffLimitMillis?: number;
 }
 
 // Interface for an initialized config.
@@ -95,7 +112,10 @@ export interface ProfilerConfig extends AuthenticationConfig {
   timeIntervalMicros: number;
   heapIntervalBytes: number;
   heapMaxStackDepth: number;
-  backoffMillis: number;
+  initialBackoffMillis: number;
+  maxBackoffMillis: number;
+  backoffMultiplier: number;
+  backoffLimitMillis: number;
 }
 
 // Default values for configuration for a profiler.
@@ -107,5 +127,8 @@ export const defaultConfig = {
   timeIntervalMicros: 1000,
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
-  backoffMillis: 5 * 60 * 1000
+  initialBackoffMillis: 1000,
+  maxBackoffMillis: parseDuration('1h'),
+  backoffMultiplier: 1.3,
+  backoffLimitMillis: parseDuration('7d'),  // 7 days
 };
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -168,10 +168,8 @@ export class Retryer {
   }
   getBackoff(): number {
     const curBackoff = Math.random() * this.nextBackoffMillis;
-    this.nextBackoffMillis = this.backoffMultiplier * this.nextBackoffMillis;
-    if (this.nextBackoffMillis > this.maxBackoffMillis) {
-      this.nextBackoffMillis = this.maxBackoffMillis;
-    }
+    this.nextBackoffMillis = Math.min(
+        this.backoffMultiplier * this.nextBackoffMillis, this.maxBackoffMillis);
     return curBackoff;
   }
   reset() {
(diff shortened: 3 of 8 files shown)


ts/src/config.ts
```diff
@@ -77,9 +78,25 @@ export interface Config extends AuthenticationConfig {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
-  // Time to wait before trying to create a profile again if profile creation
-  // fails.
-  backoffMillis?: number;
+  // On first error during profile creation, if the backoff is not specified
+  // by the server response, then profiler will between 0 and
```
aalexand: "profiler will between"?

ts/src/config.ts
```diff
@@ -77,9 +78,25 @@ export interface Config extends AuthenticationConfig {
...
+  // initialExpBackoffMillis.
+  initialBackoffMillis?: number;
+
+  // If the backoff is not specified by the server response, then profiler will
+  // wait at most maxBackoffMillis before asking server to create a profile
+  // again.
+  maxBackoffMillis?: number;
+
+  // On each consecutive error in profile creation, the maximum backoff will
+  // increase by this factor. The backoff will be random value selected
```
aalexand: "will be _a_ random value"?

ts/src/config.ts
```diff
@@ -77,9 +78,25 @@ export interface Config extends AuthenticationConfig {
...
+  // After a successful profile creation, the backoff will be reset to
+  // initialExpBackoffMillis.
+  initialBackoffMillis?: number;
+
+  // If the backoff is not specified by the server response, then profiler will
+  // wait at most maxBackoffMillis before asking server to create a profile
+  // again.
+  maxBackoffMillis?: number;
+
+  // On each consecutive error in profile creation, the maximum backoff will
```
aalexand: The use of "maximum" term is confusing here given the maxBackoffMillis name.

ts/src/config.ts
```diff
@@ -77,9 +78,25 @@ export interface Config extends AuthenticationConfig {
...
+  // increase by this factor. The backoff will be random value selected
+  // from a uniform distribution between 0 and the maximum backoff.
```
aalexand: Same.

ts/src/config.ts
```diff
@@ -77,9 +78,25 @@ export interface Config extends AuthenticationConfig {
...
+  backoffMultiplier?: number;
+
+  // Server-specified backoffs will be capped at backoffLimitMillis.
+  backoffLimitMillis?: number;
```
aalexand: Document why it's capped. We don't cap the server instructed backoff for other agents.

I also find maxBackoffMillis and backoffLimitMillis names unnecessarily inconsistent and not super clear.
* You end up using three terms here: "max", "limit" and "cap" to mean the same thing. You also sometimes use "max" to mean the current value of the backoff before the randomization applies which may confuse the reader even more, see comment above.
* The names do not make it clear that the first thing is about capping the backoff on errors and the second thing is about capping the server-instructed backoff.

I would use the "cap" term and clarify the naming / terminology here and in the code (variable names).

ts/src/profiler.ts
```diff
@@ -72,6 +73,21 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
+interface ServerBackoffResponse {
```
aalexand: Document that this corresponds to the server response carrying RetryInfo from https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto. Can we use the proto definition from that repo instead of defining the structure manually?

ts/src/profiler.ts
```diff
@@ -102,7 +118,7 @@ function isRequestProfile(prof: any): prof is RequestProfile {
 }
 
 /**
- * Returns true if response has statusCode.
+ * @return - true if response has statusCode and false otherwise.
```
aalexand: You could just say "true iff the response has the status field.", that implies the false behavior.

Do not use " - " after @return. So just "@return true iff response has the status field." Please fix the style in other places.

ts/src/profiler.ts
```diff
@@ -72,6 +73,21 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
+interface ServerBackoffResponse {
+  statusMessage: string;
+  body: {details: {retryDelay: string}};
+}
+
+/**
+ * Returns true if response indicates a backoff.
```
aalexand: Sometimes you say "Returns foo" and sometimes "@return foo". Is there a reason?

ts/src/profiler.ts
```diff
@@ -125,6 +141,50 @@ async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
...
+  // tslint:disable-next-line: no-any
+  return typeof (err as any).backoffMillis === 'number';
+}
+
+/**
+ * Class which tracks how long to wait before the next retry and can be
+ * used to get this backoff.
+ */
+export class Retryer {
+  private nextMaxBackoffMillis: number;
```
aalexand: I'd just name it "nextBackoffMillis". It's understood that randomization may change the actual value used. Sometimes randomization is implemented by randomizing the value _around_ the target value, so saying "max" wouldn't always apply anyway.

ts/src/profiler.ts
```diff
@@ -125,6 +141,50 @@ async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
...
+  constructor(
+      readonly initialBackoffMillis: number, readonly maxBackoffMillis: number,
+      readonly backoffMultiplier: number) {
+    this.nextMaxBackoffMillis = this.initialBackoffMillis;
+  }
+  getBackoff(): number {
+    const curBackoff = Math.random() * this.nextMaxBackoffMillis;
+    this.nextMaxBackoffMillis =
+        this.backoffMultiplier * this.nextMaxBackoffMillis;
+    if (this.nextMaxBackoffMillis > this.maxBackoffMillis) {
```
aalexand: Just `this.nextBackoffMillis = Math.max(this.nextBackoffMillis, this.maxBackoffMillis);`

aalexand: As I mentioned elsewhere, the usage of "max" term with two completely different meanings is extremely confusing in terms of code readability here.

ts/src/profiler.ts
```diff
@@ -216,21 +279,20 @@ export class Profiler extends common.ServiceObject {
...
-   * should be retried or with exponential backoff (up to one hour) if the
-   * response does not indicate when to retry this request.
+   * collecting another profile, or retryer specifying exponential backoff.
    */
   async collectProfile(): Promise<number> {
     let prof: RequestProfile;
     try {
       prof = await this.createProfile();
     } catch (err) {
       this.logger.error(`Failed to create profile: ${err}`);
```
aalexand: Do we log errors by default? Let's make sure we don't spam the output of the user's program. If we log errors, I think the current code will print out the error when the error is the server-instructed backoff error which is a normal condition and should not be visible to the user by default.

ts/src/profiler.ts
```diff
@@ -268,25 +330,31 @@ export class Profiler extends common.ServiceObject {
...
+          options,
+          (err: Error, prof: object, response: http.ServerResponse) => {
+            if (response && isErrorResponseStatusCode(response.statusCode)) {
+              if (isServerBackoffResponse(response)) {
+                throw new BackoffResponseError(response);
+              }
+              throw new Error(response.statusMessage);
+            }
+            if (err) {
+              reject(err);
```
aalexand: You use `reject(err)` here but `throw new SomeError()` a couple lines above - is there a difference?

ts/src/profiler.ts
```diff
@@ -268,25 +330,31 @@ export class Profiler extends common.ServiceObject {
...
-    }
-    if (!isRequestProfile(prof)) {
-      throw new Error(`Profile not valid: ${JSON.stringify(prof)}.`);
-    }
-    this.logger.debug(`Successfully created profile ${prof.profileType}.`);
-    return prof;
+    return new Promise<RequestProfile>((resolve, reject) => {
+      this.request(
+          options,
+          (err: Error, prof: object, response: http.ServerResponse) => {
```
aalexand: It might make sense to make this code fragment a separate function like `responseToProfileOrError`.

ts/src/profiler.ts
```diff
@@ -268,25 +330,31 @@ export class Profiler extends common.ServiceObject {
...
+              reject(err);
+              return;
+            }
+            if (isRequestProfile(prof)) {
+              this.logger.debug(
+                  `Successfully created profile ${prof.profileType}.`);
+              resolve(prof);
+              return;
+            }
+            throw new Error(`Profile not valid: ${JSON.stringify(prof)}.`);
```
aalexand: Not covered by tests, cover?

ts/test/test-init-config.ts
```diff
@@ -42,6 +43,16 @@ describe('initConfig', () => {
...
   });
 
+  const internalConfigParams = {
+    timeIntervalMicros: 1000,
+    heapIntervalBytes: 512 * 1024,
+    heapMaxStackDepth: 64,
+    initialBackoffMillis: 1000,
+    maxBackoffMillis: 60 * 60 * 1000,
+    backoffMultiplier: 1.3,
+    backoffLimitMillis: 7 * 24 * 60 * 60 * 1000
```
aalexand: Use parseDuration here and for other fields?

ts/src/config.ts
nolanmar511: Fixed.

ts/src/config.ts
nolanmar511: Switched to using "backoff envelope" here.

ts/src/config.ts
nolanmar511: done.

ts/src/config.ts
nolanmar511: Switched to using "backoff envelope" here.

ts/src/profiler.ts
nolanmar511: Filed an issue -- I'd need to look more into it, and if it works, I think it would be best as a separate PR.

ts/src/profiler.ts
nolanmar511: Using 'cap' instead of 'max', and clarified variable names.

ts/src/profiler.ts
nolanmar511: I should use @return everywhere. 
I've fixed this everywhere in PR #78, and am changing in profiler.ts here.

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: I've switched things logged at the error level to being logged at the warn level, and when the err is a server-instructed backoff error, I switched to logging at the debug level.

ts/src/profiler.ts
nolanmar511: There somewhat similar -- but I've switched to reject, for consistency and because it seems like there are certain cases where thrown errors can be lost.

ts/src/profiler.ts
nolanmar511: I should use "@return foo". Switching to that, and added PR #78 to fix throughout.

ts/test/test-init-config.ts
nolanmar511: I want to avoid using it here parseDuration in tests for initConfig to kind of confirm that parseDuration is working as expected.

ts/src/config.ts
nolanmar511: done.

ts/src/profiler.ts

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: Covering in PR#74 -- seems more related.

nolanmar511:
PTAL

ts/src/config.ts
```diff
@@ -77,9 +78,28 @@ export interface Config extends AuthenticationConfig {
...
+  // On each consecutive error in profile creation, the backoff envelope will
+  // increase by this factor. The backoff will be a random value selected
+  // from a uniform distribution between 0 and the backoff envelope.
+  backoffMultiplier?: number;
+
+  // On first error during profile creation, if the backoff is not specified
+  // by the server response, then profiler will wait between 0 and
+  // initialBackoffMillis before asking the server to create a profile again.
+  // After a successful profile creation, the backoff envelope will be reset to
+  // initialExpBackoffMillis.
```
aalexand: Rename to the new name.

ts/src/config.ts
```diff
@@ -77,9 +78,28 @@ export interface Config extends AuthenticationConfig {
...
+  initialBackoffMillis?: number;
+
+  // If the backoff is not specified by the server response, then profiler will
+  // wait at most expBackoffMillisCap before asking server to create a profile
+  // again.
+  expBackoffMillisCap?: number;
```
aalexand: errorBackoffCapMillis: units should go last, "exp" is not the purpose, handling errors is the purpose.

ts/src/config.ts
```diff
@@ -77,9 +78,28 @@ export interface Config extends AuthenticationConfig {
...
+
+  // Server-specified backoffs will be capped at serverBackoffMillisCap.
+  // The backoff is capped here because setTimeout (which is used to control
+  // when next profile is collected) will run immediately if the backoff is
+  // to large.
+  serverBackoffMillisCap?: number;
```
aalexand: Unit goes last in the name: serverBackoffCapMillis

ts/src/config.ts
aalexand: I would actually name it just backoffCapMillis to make consistent with backoffMultiplier and initialBackoffMillis which do not carry the prefix. So, would only use the "server" prefix for configuration related to the server instructed backoff.

ts/src/config.ts
```diff
@@ -106,5 +129,8 @@ export const defaultConfig = {
   timeIntervalMicros: 1000,
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
-  backoffMillis: 5 * 60 * 1000
+  initialBackoffMillis: 1000,
+  expBackoffMillisCap: parseDuration('1h'),
+  backoffMultiplier: 1.3,
+  serverBackoffMillisCap: parseDuration('7d'),  // 7 days
```
aalexand: The "// 7 days" comment is now redundant.

ts/src/profiler.ts
```diff
@@ -120,6 +142,69 @@ async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
   return gzBuf.toString('base64');
 }
 
+/**
+ * Error constructed from http server response which indicates backoff.
```
aalexand: Nit: http -> HTTP

ts/src/profiler.ts
```diff
@@ -120,6 +142,69 @@ async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
...
+}
+
+/**
+ * Class which tracks how long to wait before the next retry and can be
+ * used to get this backoff.
+ */
+export class Retryer {
+  private nextBackoffMillis: number;
+  constructor(
+      readonly initialBackoffMillis: number, readonly maxBackoffMillis: number,
```
aalexand: maxBackoffMillis -> backoffCapMillis since that's the terminology you use now?

ts/src/config.ts

ts/src/config.ts

ts/src/config.ts

ts/src/config.ts

ts/src/profiler.ts

ts/src/profiler.ts
nolanmar511: Yes -- Done.

ts/src/profiler.ts
```diff
@@ -212,24 +300,27 @@ export class Profiler extends common.ServiceObject {
...
+      this.logger.warn(`Failed to create profile, waiting ${
+          msToStr(backoff)} to try again: ${err}`);
+      return backoff;
     }
+    this.retryer.reset();
     try {
       await this.profileAndUpload(prof);
     } catch (err) {
-      this.logger.error(`Failed to collect and upload profile: ${err}`);
+      this.logger.warn(`Failed to collect and upload profile: ${err}`);
```
ofrobots: nit: s/collect and upload/collect or upload/? It would be nice if we could specify what failed precisely. Perhaps that can be fixed by adding logging statements inside `profileAndUpload`.

ts/src/profiler.ts
```diff
@@ -97,7 +119,7 @@ function isRequestProfile(prof: any): prof is RequestProfile {
 }
 
 /**
- * @return true if response has statusCode.
+ * @return true iff response has statusCode.
  */
 // tslint:disable-next-line: no-any
 function hasHttpStatusCode(response: any):
```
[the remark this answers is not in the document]
nolanmar511: I think using `any` makes sense here, because `hasHttpStatusCode` is type-guarding to confirm response has the fields of http.ServerResponse that are used. This is necessary, because the type of response was lost since it was returned in an array of objects. 
I have slightly updated this type-guarding, though.

ts/src/profiler.ts
```diff
@@ -14,18 +14,21 @@
...
 import {perftools} from '../../proto/profile';
 import {Common, Logger, Service, ServiceObject} from '../third_party/types/common-types';
+
 import {ProfilerConfig} from './config';
 import {HeapProfiler} from './profilers/heap-profiler';
 import {TimeProfiler} from './profilers/time-profiler';
 
 export const common: Common = require('@google-cloud/common');
 const parseDuration: (str: string) => number = require('parse-duration');
+const msToStr: (ms: number) => string = require('pretty-ms');
```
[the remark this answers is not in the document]

ts/src/config.ts
```diff
@@ -77,9 +78,28 @@ export interface Config extends AuthenticationConfig {
...
+
+  // If the backoff is not specified by the server response, then profiler will
+  // wait at most backoffCapMillis before asking server to create a profile
+  // again.
+  backoffCapMillis?: number;
+
+  // Server-specified backoffs will be capped at serverBackoffCapMillis.
+  // The backoff is capped here because setTimeout (which is used to control
+  // when next profile is collected) will run immediately if the backoff is
+  // to large.
```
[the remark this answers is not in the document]
Referenced the issue and switched to using 2147483646 as serverBackoffCap. (I did one less that the value you noted, just to be a bit careful).

ts/src/profiler.ts
```diff
@@ -72,6 +75,25 @@ export interface RequestProfile {
...
+ */
+// tslint:disable-next-line: no-any
+function getServerBackoffResponse(response: any): number|undefined {
+  if (response.body && response.body.error && response.body.error.details &&
+      response.body.error.details instanceof Array) {
+    for (let i = 0; i < response.body.error.details.length; i++) {
+      const item = response.body.error.details[i];
+      if (typeof item === 'object' && item.retryDelay &&
+          typeof item.retryDelay === 'string') {
+        return parseDuration(item.retryDelay);
```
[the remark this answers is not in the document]
nolanmar511: How I handled this didn't match the comments, so I've updated this.
Done.

ts/src/profiler.ts
```diff
@@ -120,6 +142,69 @@ async function profileBytes(p: perftools.profiles.IProfile): Promise<string> {
...
+    super(response.statusMessage);
+  }
+}
+
+/**
+ * @return true if error is a BackoffResponseError and false otherwise
+ */
+function isBackoffResponseError(err: Error): err is BackoffResponseError {
+  // tslint:disable-next-line: no-any
+  return typeof (err as any).backoffMillis === 'number';
```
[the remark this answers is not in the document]

ts/src/profiler.ts
nolanmar511: I've shifted the logging to profileAndUpload. 
Hopefully code is still clear.

nolanmar511:
PTAL

ts/src/config.ts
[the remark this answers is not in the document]
aalexand: I don't think extra care is needed, the doc says "When delay is larger than 2147483647 or less than 1, the delay will be set to 1." so you can just use 2147483647. By the way, where do you use the new value, I can still see the default config using parseDuration('7d')?

ts/src/config.ts
```diff
@@ -106,5 +130,8 @@ export const defaultConfig = {
   timeIntervalMicros: 1000,
   heapIntervalBytes: 512 * 1024,
   heapMaxStackDepth: 64,
-  backoffMillis: 5 * 60 * 1000
+  initialBackoffMillis: 1000,
+  backoffCapMillis: parseDuration('1h'),
+  backoffMultiplier: 1.3,
+  serverBackoffCapMillis: 2147483647
```
aalexand: I'd add the doc link here as well to indicate where this value comes from.

nolanmar511:
@ofrobots  -- any additional comments?

aalexand:
@ofrobots ping

ts/src/profiler.ts
```diff
@@ -72,6 +75,29 @@ export interface RequestProfile {
...
+ */
+// tslint:disable-next-line: no-any
+function getServerResponseBackoff(response: any): number|undefined {
+  if (response && response.body && response.body.error &&
+      response.body.error.details &&
+      response.body.error.details instanceof Array) {
+    for (let i = 0; i < response.body.error.details.length; i++) {
+      const item = response.body.error.details[i];
+      if (typeof item === 'object' && item.retryDelay &&
+          typeof item.retryDelay === 'string') {
```
ofrobots: Same comment about structurally malformed responses from the service. Should we log?

package.json
```diff
@@ -26,13 +26,15 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@google-cloud/common": "^0.15.0",
+    "@types/pretty-ms": "^3.0.0",
```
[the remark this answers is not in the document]
nolanmar511: Yes -- done.

ts/src/profiler.ts
```diff
@@ -72,6 +75,29 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
+/**
+ * @return number indicated by backoff if the response indicates a backoff and
+ * that backoff is greater than 0. Otherwise returns undefined.
+ */
+// tslint:disable-next-line: no-any
+function getServerResponseBackoff(response: any): number|undefined {
```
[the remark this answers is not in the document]
nolanmar511: I did this because http.ServerResponse doesn't have a field  body. So if I use http.ServerResponse as the type, then I would have to cast the response to any throughout this function.
Would that be preferable?

ts/src/profiler.ts
```diff
@@ -72,6 +75,29 @@ export interface RequestProfile {
...
+  if (response && response.body && response.body.error &&
+      response.body.error.details &&
+      response.body.error.details instanceof Array) {
```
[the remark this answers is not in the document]
nolanmar511: I'm not sure that logging a warning whenever a backoff is not specified in this format makes sense, since a server-specified backoff is only expected when 

I do log the backoff [here](https://github.com/nolanmar511/cloud-profiler-nodejs/blob/1fc8cbf858618db282a35239d649125e37e30154/ts/src/profiler.ts#L313), and log errors differently depending on whether or not the error has a backoff. Would this be sufficient for observing if there is a problem with getting the backoff?

ts/src/profiler.ts
```diff
@@ -72,6 +75,29 @@ export interface RequestProfile {
...
+    for (let i = 0; i < response.body.error.details.length; i++) {
```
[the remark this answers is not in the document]

ts/src/profiler.ts
nolanmar511: Would this logging be used to indicate how the response is malformed?
If so, it's pretty easy to reproduce the error which we would expect to specify a backoff (just start about 20 copies of a trivial application with matching configurations and wait), which would make figuring out how something is malformed less important.

ts/src/profiler.ts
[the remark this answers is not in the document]
nolanmar511: I've switched to using http.ServerResponse as the type.

ts/src/profiler.ts
```diff
@@ -72,6 +75,29 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
+/**
+ * @return number indicated by backoff if the response indicates a backoff and
+ * that backoff is greater than 0. Otherwise returns undefined.
+ */
+// tslint:disable-next-line: no-any
+function getServerResponseBackoff(response: any): number|undefined {
+  if (response && response.body && response.body.error &&
```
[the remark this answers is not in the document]

ts/src/profiler.ts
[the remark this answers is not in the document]

nolanmar511:
PTAL

ts/src/profiler.ts
```diff
@@ -401,9 +392,8 @@ export class Profiler extends common.ServiceObject {
 
     try {
       const [body, response] = await this.request(options);
-      if (!hasHttpStatusCode(response)) {
-        this.logger.debug(
-            'Server response missing status information when attempting to upload profile.');
+      if (!(response instanceof http.ServerResponse)) {
```
ofrobots: `instanceof` isn't idiomatic and safe to use in JavaScript where typing is structural rather than nominal. See http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/ for some background.

ts/src/profiler.ts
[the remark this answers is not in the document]
ofrobots: No that log doesn't really address my concerns. I am not worried about the backoff but rather the weak typing in JavaScript. There is no easy way for me, as a code reviewer, to verify that the code here matches what actual real server response looks like.

I would be less concerned if this was build on top of stronger types - say provided by #87. I would be okay if you want to address this as a follow-on.

ts/src/profiler.ts
ofrobots: > If so, it's pretty easy to reproduce the error which we would expect to specify a backoff

That sounds like e2e test. Excellent! Let's add one.

ts/src/profiler.ts
nolanmar511: Done -- I've stopped using instanceof.

nolanmar511:
I have stopped using instanceof. 
I have manually tested this PR. And I will be working on an e2e test (this is my next priority

nolanmar511:
PTAL

PR #76, opened 2017-11-22
fix: fix ESOCKETTIMEDOUT Error by increasing timeout to 1 hour for createProfile()

### LINKED ISSUE #70: Sporadic ESOCKETTIMEDOUT error during profiling
I am running the following program on GCE:
```
require('@google-cloud/profiler').start({
    serviceContext: {
    service: 'test-service',
    version: '1.0.0',
    logLevel: 4,
  },
});

var doSomeWork = () => {
  console.log("doing some work");
  for (var i = 0; i < 1e8; i++) {
  }
};

setInterval(doSomeWork, 1000);
```

Once in a while I am getting an error like:
```
...
doing some work
doing some work
ERROR:@google-cloud/profiler: Error requesting profile type to be collected: Error: ESOCKETTIMEDOUT.
doing some work
doing some work
...
```

There may be some unexpected timeout happening on the profile creation.

nolanmar511:
Fixes #70 

I tested by running 8 copies of the acmeAir app both before and after this change (for about 10 minutes). Before, ESOCKETTIMEDOUT errors occurred on every request to createProfile(), after, no ESOCKETTIMEDOUT errors were observed.

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -268,10 +268,10 @@ export class Profiler extends common.ServiceObject {
       body: reqBody,
       json: true,
 
-      // Default timeout for for request is 1 minute, but request to create
+      // Default timeout for for a request is 1 minute, but request to create
       // profile is designed to hang until it is time to collect a profile
       // (up to one hour).
-      timeout: 60 * 60 * 1000,
+      timeout: parseDuration('1h'),
     };
 
     this.logger.debug(`Attempting to create profile.`);
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -267,6 +267,7 @@ export class Profiler extends common.ServiceObject {
       uri: '/profiles',
       body: reqBody,
       json: true,
+      timeout: 60 * 60 * 1000,
     };
 
     this.logger.debug(`Attempting to create profile.`);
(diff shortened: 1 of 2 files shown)


ts/src/profiler.ts
```diff
@@ -267,6 +267,7 @@ export class Profiler extends common.ServiceObject {
       uri: '/profiles',
       body: reqBody,
       json: true,
+      timeout: 60 * 60 * 1000,
```
aalexand: It would deserve a comment about why the timeout is overridden from the default.

aalexand: You don't have new tests in the PR, so presumably some manual testing was done to verify the fix. Can you describe the testing you did in the PR description?

ts/src/profiler.ts
```diff
@@ -267,6 +267,11 @@ export class Profiler extends common.ServiceObject {
       uri: '/profiles',
       body: reqBody,
       json: true,
+
+      // Default timeout for for request is 1 minute, but request to create
+      // profile is designed to hang until it is time to collect a profile
+      // (up to one hour).
+      timeout: 60 * 60 * 1000,
```
aalexand: Also: I thought we started to use parseDuration('1h')-like expressions instead of doing the direct math with millis?

nolanmar511: Yes -- Done.

gaofanmichael:
Looks like if you delete empty line 270 in ts/test/test-profiler.ts, gts will be happy.

nolanmar511:
Tests working now -- PTAL

ofrobots:
Please use https://conventionalcommits.org/ style for commit messages.

PR #78, opened 2017-11-22
fix: replace 'Returns' in comments with '@return'

nolanmar511:
Using @return is correct style for nodejs.


ofrobots:
Please use https://conventionalcommits.org/ style for commit messages.

aalexand:
@ofrobots Hmm, not sure why the "fix: " prefix didn't make it to the commit message.

PR #85, opened 2017-11-29
chore: add testing for osx

nolanmar511:
Have not yet confirmed that changes work.
Will update this description and add a PTAL when ready.


nolanmar511:
Opened pull request to check if testing with travis on OSX might be faster with the PR, it was not. So closing this PR.

Issue #77, opened 2017-11-22
Use type definitions from error_details.proto


nolanmar511:
ServerBackoffResponse (in pr #75) corresponds to a server response with RetryInfo  in error_details.proto (here https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto). We should use (or at least investigate using) error_details.proto.

aalexand:
Whose next step here: @nolanmar511 or @ofrobots ? The issue is assigned to @ofrobots, is that right?

nolanmar511:
It should be assigned to me. 
I need to check if it is reasonable to switch to using the error_details definition.

PR #87, opened 2017-11-30
build: add error_details definitions

ofrobots:
~~DO NOT LAND YET~~

Add js & type definitions generated from error_details.proto.

Fixes: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/issues/77

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "prepare": "npm run compile",
     "pretest": "npm run compile",
     "posttest": "npm run check",
-    "proto": "npm run proto:profile && npm run proto:profiler",
+    "proto": "npm run proto:profile && npm run proto:profiler && npm run proto:error-details",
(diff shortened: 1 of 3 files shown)


aalexand:
This doesn't fix #77 alone, right? I expect we should also start using it. Or am I commenting against unfinished work? Apologies if so.

ofrobots:
Well, since #75 hasn't landed yet, we could either consider this to 'fix' #77, and adjust #75, or wait for #75 to land and address the issue here.

I would be okay with either.

aalexand:
Ah, OK. I'm fine with either too.

aalexand:
@nolanmar511 As discussed, take these changes and make the definitions actually used, as a separate PR. This one will be closed then.

ofrobots:
I would like to close this PR at this point. It would be fairly easy to get back the definitions, and the issue #77 remains open to track the work needed.

If this is a mistake, go ahead and reopen.

PR #88, opened 2017-12-01
feat: add way to profile locally

nolanmar511:
Adds a function which can be used to collect profiles locally. This is intended for debugging.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -125,28 +125,24 @@ export async function start(config: Config = {}): Promise<void> {
 
 
 /**
- * For debugging purposes. Collects profiles and writes profiles
- * out to console in the format used to upload the profiles to the API.
+ * For debugging purposes. Collects profiles and discards the collected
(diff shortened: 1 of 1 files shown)


ts/src/index.ts
```diff
@@ -122,6 +123,35 @@ export async function start(config: Config = {}): Promise<void> {
...
+  const normalizedConfig = await initConfig(config);
+  profiler = new Profiler(normalizedConfig);
+  
+  while (true) {
+    if(!config.disableHeap) {
+      const heap = await profiler.profile({
+        name: 'HEAP-Profile' + new Date(),
+        profileType: 'HEAP'
+      });
+      console.log(heap);
```
aalexand: Will it print binary content to the console? That's crude if so. I'd either discard the collected profiles or save to temporary files.

nolanmar511: I'm going to discard the collected profiles.

PR #93, opened 2017-12-19
fix: set default log level to warn

### LINKED ISSUE #82: Change the default log level for "warning" and adhere that in the source
And further:
* The default level should be “warning”.
* Use error level for only for invalid profiler configuration by user. Should not be more than a couple in the whole source of the agent.
* Use warning level for persistent / unexpected errors which user should know about because they degrade the profiling quality. Should be used very sparingly.
* The info and debug levels should be used for logging information that can be useful to investigate issues. Getting this will require users to change the default log level per maintainer request.

nolanmar511:
Fixes #82

### DIFF
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -124,7 +124,7 @@ export interface ProfilerConfig extends AuthenticationConfig {
 
 // Default values for configuration for a profiler.
 export const defaultConfig = {
-  logLevel: 1,
+  logLevel: 2,
   serviceContext: {},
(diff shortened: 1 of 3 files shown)


aalexand:
As we discussed, this does not meet all the rules listed in #82?

nolanmar511:
Updated to meet all conditions of #82.
Specifically, added a logging statement at the error level if the configuration cannot be initialized and switched the default log level to 2 (the warning level).

PR #95, opened 2017-12-21
Save profiles when running startLocal

nolanmar511:
Right now, profiles are thrown away when running startLocal. With this change, profiles will be written out to files.


nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -148,13 +151,27 @@ export async function startLocal(config: Config = {}): Promise<void> {
     if (!config.disableHeap) {
       const heap = await profiler.profile(
           {name: 'HEAP-Profile' + new Date(), profileType: 'HEAP'});
+      const decodedBytes = Buffer.from(heap.profileBytes as 'string', 'base64');
+      fs.writeFile(
```
aalexand: Would be good to document somewhere in the code where the profiles will get written (current directory? temp directory?)

nolanmar511: Done

aalexand:
Should this PR be closed or submitted?

nolanmar511:
I think I'll close this. If startLocal does save profiles, it should probably be only when the user specifies this.

PR #98, opened 2018-01-02
fix: remove clang-format dependency

nolanmar511:
fixes #97

PR #99, opened 2018-01-02
chore: update gts dependency

nolanmar511:
Based on discussion in #98, we decided to update gts dependency so that the version of gts used would depend on a fixed version of clang-format.


aalexand:
Node.js 6 CI build seems broken?

nolanmar511:
PR #98 should fix failing tests on node 6.

PR #101, opened 2018-01-05
fix: fix incorrect time profile duration

### LINKED ISSUE #100: Incorrect profile duration for time profiles
The recorded duration for a time profile is around 10000 seconds, so 1000 times greater than it should be.

nolanmar511:
fixes #100


aalexand:
Travis is unhappy?

PR #102, opened 2018-01-06
fix: fix travis tests on node 6

nolanmar511:
Travis tests with node 6 are failing with:
```
ts/src/index.ts(76,30): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
76         require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
```

This PR adds a cast to process.env.GCLOUD_PROFILER_CONFIG, so tests again pass on node 6.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -71,9 +71,9 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   }
 
   let envSetConfig: Config = {};
-  if (process.env.hasOwnProperty('GCLOUD_PROFILER_CONFIG')) {
-    envSetConfig =
-        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
+  let val =  require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG!)) as Config;
+  if (val) {
+    envSetConfig = val;
   }
 
   const mergedConfig =
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/profiler",
-  "version": "0.1.9",
+  "version": "0.1.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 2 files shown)


ts/src/index.ts
```diff
@@ -72,8 +72,12 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
 
   let envSetConfig: Config = {};
   if (process.env.hasOwnProperty('GCLOUD_PROFILER_CONFIG')) {
+    // Type of process.env.GCLOUD_PROFILER_CONFIG is string | undefined
+    // but we know that it is defined.
+    // TODO: figure out how to avoid this cast.
```
aalexand: Couldn't you just do
```
let val = process.env.GCLOUD_PROFILER_CONFIG;
if (val) {
  envSetConfig = require(path.resolve(val)) as Config;
}
```
or something like that?

ts/src/index.ts
```diff
@@ -72,8 +72,12 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
     envSetConfig =
-        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
+        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG as string)) as
```
kjin: Try `process.env.GCLOUD_PROFILER_CONFIG!`, which coerces it to a non-null type. We should use it sparingly but in this case it's fine because we're already calling `hasOwnProperty` immediately beforehand

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -72,8 +72,10 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
 
   let envSetConfig: Config = {};
   if (process.env.hasOwnProperty('GCLOUD_PROFILER_CONFIG')) {
+    // Type of process.env.GCLOUD_PROFILER_CONFIG is string | undefined
+    // but we know that it is defined.
     envSetConfig =
-        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
+        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG!)) as Config;
```
aalexand: Why not assign first, then check like I suggested earlier? Or that won't work?

nolanmar511:
Tests on 6 are once again passing. I'm closing this.

Issue #68, opened 2017-11-20
Add back source-map-support for better stack traces 


nolanmar511:
When running tests in #56, tests passed prior to merging in #60 and failed after, with the following stack trace:

```
/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/util.js:344
  var cmp = mappingA.generatedLine - mappingB.generatedLine;
                                              ^

TypeError: Cannot read property 'generatedLine' of undefined
    at compareByGeneratedPositionsDeflated (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/util.js:344:47)
    at doQuickSort (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/quick-sort.js:88:11)
    at exports.quickSort (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/quick-sort.js:113:3)
    at BasicSourceMapConsumer.SourceMapConsumer_parseMappings [as _parseMappings] (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/source-map-consumer.js:564:5)
    at BasicSourceMapConsumer.get [as _generatedMappings] (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/source-map-consumer.js:70:12)
    at BasicSourceMapConsumer.SourceMapConsumer_originalPositionFor [as originalPositionFor] (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map/lib/source-map-consumer.js:655:12)
    at mapSourcePosition (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map-support/source-map-support.js:199:42)
    at wrapCallSite (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map-support/source-map-support.js:343:20)
    at /usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map-support/source-map-support.js:378:26
    at Array.map (<anonymous>)
    at Function.prepareStackTrace (/usr/local/google/home/nolanmar/cloud-profiler-nodejs/node_modules/source-map-support/source-map-support.js:377:24)
[...]

PR #107, opened 2018-01-09
feat: add source mapping to tests for better stack traces

nolanmar511:
fixes: #68 

Tests with source mapping has previously failed when Math.random() was stubbed with sinon. This adds source-mapping back and modifies code so that Math.random() does not need to be stubbed.

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -159,13 +159,18 @@ function isBackoffResponseError(err: Error): err is BackoffResponseError {
  */
 export class Retryer {
   private nextBackoffMillis: number;
+
+  // For testing. Allows Math.random() to be replaced with non-random function.
+  private random: () => number;
+
   constructor(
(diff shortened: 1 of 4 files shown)


ts/src/profiler.ts
```diff
@@ -159,13 +159,18 @@ function isBackoffResponseError(err: Error): err is BackoffResponseError {
...
 export class Retryer {
   private nextBackoffMillis: number;
+
+  // For testing. Allows Math.random() to be replaced with non-random function.
+  private random: () => number;
+
   constructor(
       readonly initialBackoffMillis: number, readonly backoffCapMillis: number,
-      readonly backoffMultiplier: number) {
+      readonly backoffMultiplier: number, random = Math.random) {
```
aalexand: Is this Math.random change related to the source map fix? It is not obvious that it is.

nolanmar511: Yes -- adding a stub of Math.random() was the change which had previously caused tests with source maps to fail.

PR #108, opened 2018-01-09
chore: Update @types/node to the latest version

nolanmar511:
Greenkeeper's PR #104 for updating @types/node to 9.3.0 has failing tests. 
This change updates @types/node and modified code so tests continue to pass (specifically ts/src/index.ts has a change).

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -71,9 +71,9 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   }
 
   let envSetConfig: Config = {};
-  if (process.env.GCLOUD_PROFILER_CONFIG) {
-    envSetConfig =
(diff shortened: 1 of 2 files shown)


ts/src/index.ts
```diff
@@ -71,9 +71,9 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   }
 
   let envSetConfig: Config = {};
-  if (process.env.GCLOUD_PROFILER_CONFIG) {
-    envSetConfig =
-        require(path.resolve(process.env.GCLOUD_PROFILER_CONFIG)) as Config;
+  const val = process.env.GCLOUD_PROFILER_CONFIG;
```
aalexand: Should this change be part of this PR?

nolanmar511: Yes -- when upgrading to @types/node 9.3.0, there is a type error in this section of code without a change.

Issue #115, opened 2018-01-18
Memory leak in Heap Profiler


nolanmar511:
There is a memory leak when using the profiler and collecting just time profiles, just heap profiles, or both time and heap profiles. 
Increasing the rate at which profiles are collected seems to increase the rate at which memory leaks. When using `process.memoryUsage()` to look at memory used, RSS increases notable and heapTotal seems to roughly constant.

ofrobots:
Is the leak in the JavaScript heap or in external heap?

nolanmar511:
External heap.

nolanmar511:
This has been fixed by #117.

PR #117, opened 2018-01-19
fix: use delete instead of free in sampling-heap-profiler.cc

nolanmar511:
This helps with #115.


bindings/sampling-heap-profiler.cc
```diff
@@ -77,7 +77,7 @@ NAN_METHOD(GetAllocationProfile) {
   AllocationProfile* profile = info.GetIsolate()->GetHeapProfiler()->GetAllocationProfile();
   AllocationProfile::Node* root = profile->GetRootNode();
   info.GetReturnValue().Set(TranslateAllocationProfile(root));
-  free(profile);
+  delete profile;
```
ofrobots: I would suggest automatic memory management instead.. something like:

```c++
std::unique_ptr<v8::AllocationProfile> profile(
    info.GetIsolate()->GetHeapProfiler()->GetAllocationProfiler());
```

nolanmar511: Done.

PR #127, opened 2018-02-08
chore: update gcp-metadata

nolanmar511:
Updates gcp-metadata to version 0.5.0.
Also makes the changes so that code uses gcp-metadata's new API.

In addition to running the tests, I also ran a trivial application in GCE and saw that the zone was set in the profiles after this update.


aalexand:
> In addition to running the tests, I also ran a trivial application in GCE and saw that the zone was set in the profiles after this update.

This is exactly kind of thing that we need to have end-to-end tests for.

aalexand:
nolanmar@, I can review this but it would be best if an owner of gcp-metadata package would review.

PR #131, opened 2018-02-13
feat: add field to config so user can use test API

nolanmar511:
It's useful when testing the agent to be able to upload profiles to the test API.
This PR adds a field to the config, `baseApiUrl`, which a user can use to set the api url used to something other than `cloudprofiler.googleapis.com`.

### DIFF
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -102,6 +102,10 @@ export interface Config extends AuthenticationConfig {
   // too large.
   // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
   serverBackoffCapMillis?: number;
+
+  // Use test-cloudprofiler.sandbox.googleapis.com instead of
+  // cloudprofiler.googleapis.com when this is set to true.
+  useTestApi?: boolean;
 }
 
 // Interface for an initialized config.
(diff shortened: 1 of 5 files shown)


ts/src/config.ts
```diff
@@ -102,6 +102,10 @@ export interface Config extends AuthenticationConfig {
   // too large.
   // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
   serverBackoffCapMillis?: number;
+
+  // Use test-cloudprofiler.sandbox.googleapis.com instead of
+  // cloudprofiler.googleapis.com when this is set to true.
+  useTestApi?: boolean;
```
aalexand: Maybe expose it as baseApiUrl instead so that if there are other environment we want to test in future we can use the same flag?

nolanmar511: Done!

nolanmar511:
PTAL

ts/src/config.ts
```diff
@@ -102,6 +102,9 @@ export interface Config extends AuthenticationConfig {
   // too large.
   // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
   serverBackoffCapMillis?: number;
+
+  // Allows user to specify API URL other than cloudprofiler.googleapis.com/v2.
```
aalexand: Nit: "other than https://cloudprofiler.googleapis.com/v2" so that it's verbatim of what it is and can serve as a reference.

PR #137, opened 2018-02-21
fix: update type field used for value types

nolanmar511:
Right now, the UI we're using to view profiles is having problems with unknown metric names. 
Unfortunately, this UI is slow to update.
As a work-around, we can change the type of the wall profile's value type from 'time' to 'wall'.

This has another advantage : it makes the value types reported for the wall profile more consistent with the value types profiling agents for other languages use. 

To increase consistency further, this also changes the type 'samples' to 'sample'


ts/src/profilers/profile-serializer.ts
```diff
@@ -170,7 +170,9 @@ function createSampleCountValueType(table: StringTable):
  */
 function createTimeValueType(table: StringTable): perftools.profiles.ValueType {
   return new perftools.profiles.ValueType({
-    type: table.getIndexOrAdd('time'),
+    // TODO: change type back to time.
```
aalexand: Java agent puts the time metric in the wall profiles named as "wall", so I'd remove the TODO and leave the metric name as "wall" here after all.

BTW, no test changes needed?

nolanmar511: Forgot to run the tests -- the tests have now been updated.
Removing TODO now.

ts/src/profilers/profile-serializer.ts
```diff
@@ -170,7 +170,9 @@ function createSampleCountValueType(table: StringTable):
...
+    // This is a work-around so profiles can be viewed using the UI.
+    type: table.getIndexOrAdd('wall'),
```
aalexand: I do not like this Github feature where it's not possible to comment on lines that were not changed in the PR, but several lines above you add "samples" metric. Java agent names that "sample", in singular, so let's change to "sample" here as well.

sample

ts/test/profiles-for-tests.ts
```diff
@@ -137,7 +137,7 @@ export const timeProfile: perftools.profiles.IProfile = {
     '',
     'samples',
```
aalexand: samples -> sample

nolanmar511: Done (here and elsewhere)

ts/src/profilers/profile-serializer.ts
nolanmar511: Done!

aalexand:
Looks good, but update the title and description to 1) mention the samples -> sample change; 2) have a standard commit prefix?

nolanmar511:
Mentioned samples to sample change, and added a standard commit prefix (I went with fix because we need this for things to work now; otherwise chore might have been better)

PR #141, opened 2018-02-22
fix: don't run profiler if incompatible version of node is used

nolanmar511:
This PR adds `engines` to the package.json and prevents the profiler from starting when a version of node not matching the version specified in `engines` is used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -155,6 +155,12 @@
         "@types/node": "9.4.5"
       }
     },
+    "@types/semver": {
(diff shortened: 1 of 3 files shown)


aalexand:
@nolanmar511 Looks like CI is failing.

aalexand:
I wonder why the code coverage has decreased.

nolanmar511:
Code coverage decreased because I added lines of code to `start()`, which is not tested because that function would not exit under testing conditions.

PR #142, opened 2018-03-01
chore: add benchmark which will be used for e2e testing

nolanmar511:
We need to use this benchmark for other testing of the agent, so are adding it before the rest of the integration test.

### DIFF
--- a/testing/busybench.js
+++ b/testing/busybench.js
@@ -1,0 +1,34 @@
+/**
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


testing/busybench.js
```diff
@@ -0,0 +1,14 @@
+const startTime = Date.now();
```
aalexand: Add copyright header and a bit of doc?

nolanmar511: Done.

nolanmar511:
PTAL

testing/busybench.js
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const startTime = Date.now();
+
+// benchmark does work filling a benchmark, then calls itself with setImmediate.
```
aalexand: Function name should not be repeated in JSDoc.

"does work filling a benchmark" - can't parse.

nolanmar511: Done

nolanmar511:
PTAL

PR #147, opened 2018-03-07
chore: added kokoro configurations

nolanmar511:
Add kokoro configurations.
This is a step needed to enable integration tests.


nolanmar511:
@jjqll -- could you take a look at this?

PR #148, opened 2018-03-07
fix: do not ask CPU profiler for individual samples

### LINKED ISSUE #145: Do not ask CPU profiler for individual samples as timestamps are discarded anyway?
The CPU profiler start is currently called from here: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/8b418473fc57e0f77737f8461bace4e95e98296d/bindings/time-profiler.cc#L64.

And the ts code passes true to the record_samples argument from here: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/8b418473fc57e0f77737f8461bace4e95e98296d/ts/src/profilers/time-profiler.ts#L39.

The record_samples parameter is documented here: https://github.com/nodejs/node/blob/4e15679c026d2969efa276149282df388e0533b3/deps/v8/include/v8-profiler.h#L319 as

```
   * |record_samples| parameter controls whether individual samples should
   * be recorded in addition to the aggregated tree.
```

But it appears the profiler only uses aggregated information to populate the profile since the information is accessed using the top-down root which is aggregated structure: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/8b418473fc57e0f77737f8461bace4e95e98296d/bindings/time-profiler.cc#L51.

The sample count is converted to the JS land here - https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/8b418473fc57e0f77737f8461bace4e95e98296d/bindings/time-profiler.cc#L53 but it doesn't seem to be used while converting the profile to profile.proto.

So it seems that we ask the profiler to store the individual samples but don't really need it.

@nolanmar511 @ofrobots WDYT?

nolanmar511:
After making this change, I ran the integration test (which still passed) and ran the script used for the integration test locally (modifying to start the profiler in the script) and saw that symbolized time profiles with samples were collected. 

Fixes #145


bindings/time-profiler.cc
```diff
@@ -60,8 +60,7 @@ Local<Value> TranslateTimeProfile(const CpuProfile* profile) {
 
 NAN_METHOD(StartProfiling) {
   Local<String> name = info[0].As<String>();
-  bool record_samples = info[1].As<Boolean>()->BooleanValue();
-  info.GetIsolate()->GetCpuProfiler()->StartProfiling(name, record_samples);
+  info.GetIsolate()->GetCpuProfiler()->StartProfiling(name, false);
```
aalexand: I would add a comment on what `false` here is and why `false` is good enough for us.

PR #154, opened 2018-03-08
feat: allow users to set projectId with GOOGLE_CLOUD_PROJECT

### LINKED ISSUE #152: Respect GOOGLE_CLOUD_PROJECT, not GCLOUD_PROJECT
GOOGLE_CLOUD_PROJECT is the environment variable we should use, I am pretty sure GCLOUD_PROJECT is deprecated / old. E.g. see https://cloud.google.com/appengine/docs/flexible/nodejs/runtime#environment_variables

nolanmar511:
If both GCLOUD_PROJECT and GOOGLE_CLOUD_PROJECT are set, GOOGLE_CLOUD_PROJECT will be used as the projectId. 

When making this change, I found that `common.util.normalizeArguments(null, config);` sets the projectId of the output to GCLOUD_PROJECT when no projectId is set. To allow GOOGLE_CLOUD_PROJECT to take priority over GCLOUD_PROJECT, I modified how we use `common.util.normalizeArguments()`.

Fixes #152.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -53,10 +53,10 @@ function hasService(config: Config):
  * Exported for testing purposes.
  */
 export async function initConfig(config: Config): Promise<ProfilerConfig> {
-  config = common.util.normalizeArguments(null, config);
+  const normalizedConfig: Config = common.util.normalizeArguments(null, config);
 
   const envConfig: Config = {
-    projectId: process.env.GCLOUD_PROJECT,
+    projectId: process.env.GOOGLE_CLOUD_PROJECT || process.env.GCLOUD_PROJECT,
     serviceContext: {
       service: process.env.GAE_SERVICE,
       version: process.env.GAE_VERSION,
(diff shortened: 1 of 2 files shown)


ts/src/index.ts
```diff
@@ -53,10 +53,10 @@ function hasService(config: Config):
  * Exported for testing purposes.
  */
 export async function initConfig(config: Config): Promise<ProfilerConfig> {
-  config = common.util.normalizeArguments(null, config);
+  const normalizedConfig: Config = common.util.normalizeArguments(null, config);
```
aalexand: Why is this change made? Is this related? Description says nothing about other changes, so I expected to only see the environment variable change.

nolanmar511: Thanks -- I've updated the description to explain this.

aalexand:
@nolanmar511 Can you file an issue against this common library so that it handles both GCLOUD_PROJECT and GOOGLE_CLOUD_PROJECT for us? @ofrobots FYI

aalexand:
@nolanmar511 Keeping this PR open, will approve once the issue is filed.

aalexand:
Turns out (see latest common in #152) the agent already supports GOOGLE_CLOUD_PROJECT via the common library we use. We can drop this PR, or rather probably drop the code that deals with the environment variables altogether.

nolanmar511:
Closing this. 
GOOGLE_CLOUD_PROJECT is already supported.

PR #155, opened 2018-03-08
fix: update integration test so it runs successfully.

nolanmar511:
This PR includes small changes needed to allow e2e tests to run successfully. 

Most notably, it stops using `fill()` in busybench.js -- based on profiles, it appears that `fill()` now uses a setImmediate somewhere in node 8 and node 9, so the work was no longer attributed to `benchmark()`.

### DIFF
--- a/testing/integration_test.sh
+++ b/testing/integration_test.sh
@@ -15,7 +15,7 @@ REPO=$(git config --get remote.origin.url)
 
 
 export GCLOUD_TESTS_NODEJS_PROJECT_ID="cloud-profiler-e2e"
-export GCLOUD_TESTS_NODEJS_ZONE="us-east1-a"
+export GCLOUD_TESTS_NODEJS_ZONE="us-east1-b"
 export GOOGLE_APPLICATION_CREDENTIALS="${SERVICE_KEY}"
 
 # Move test to go path.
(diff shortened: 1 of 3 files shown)


aalexand:
> fill() now uses a setImmediate somewhere in node 8 and node 9

@nolanmar511 This feels weird because that means fill() is asynchronous and upon its exit the array hasn't been yet filled?

nolanmar511:
@aalexand -- I've updated the description, though I'm having a hard time re-finding profiles from runs that gave me trouble.
A few times `benchmark()` didn't show up in profiles; this change seemed to resolve the problem.

testing/integration_test.sh
```diff
@@ -15,7 +15,7 @@ REPO=$(git config --get remote.origin.url)
 
 
 export GCLOUD_TESTS_NODEJS_PROJECT_ID="cloud-profiler-e2e"
-export GCLOUD_TESTS_NODEJS_ZONE="us-east1-a"
+export GCLOUD_TESTS_NODEJS_ZONE="us-east1-b"
```
aalexand: Why this change was done? Just curious.

aalexand:
> A few times benchmark() didn't show up in profiles; this change seemed to resolve the problem.
@nolanmar511 Seems like we don't know what exactly is going on then. I could as well see this being a bug in the profiler or runtime we've now shadowed.

testing/integration_test.sh
nolanmar511: us-east1-a doesn't exist.

nolanmar511:
Test is still flaky. I don't believe it was originally flaky (I'd not, until recently seen it fail with profile collected but the function name missing).
Current theory: the flakiness may have been introduced when setting record_samples to false. 
I'll file a bug for this.

PR #156, opened 2018-03-08
chore: Update @google-cloud/common to ^0.16.1

nolanmar511:


ts/system-test/test-start.ts
```diff
@@ -73,9 +73,9 @@ before(async () => {
       .reply(200, (request: RequestProfile, body: RequestProfile) => {
         tempUploadedProfiles.push(body);
       });
-  nock('https://accounts.google.com')
+  nock(/https:\/\/(accounts\.google\.com|www\.googleapis\.com)/)
```
ofrobots: Are both URLs possible with 0.16.1?

nolanmar511: I think only googleapis.com is needed. Done.

nolanmar511:
PTAL

PR #158, opened 2018-03-09
fix: modify benchmark so that it fills arrays, rather than a buffer

nolanmar511:
The buffer wasn't showing up too much in the heap profiles, so the integration test didn't always find the `benchmark()` function.

With this change, the benchmark() function dominates memory allocations.

### DIFF
--- a/testing/busybench.js
+++ b/testing/busybench.js
@@ -15,20 +15,26 @@
  */
 
 const startTime = Date.now();
+const testArr = [];
 
 /**
- * Repeatedly fills a buffer, then calls itself with setImmediate.
+ * Fills several arrays, then calls itself with setImmediate.
  * It continues to do this until durationSeconds after the startTime.
  */
 function benchmark(durationSeconds) {
-  var buffer = new Buffer(1e4);
-  for (var k = 0; k < 1e4; k++) {
-    for (var j = 0; j < buffer.length; j++) {
-      buffer[j] = 0;
+  for (let i = 0; i < 200; i++) {
+    testArr[i] = new Array(512 * 1024);
+    for (let j=0; j < testArr[i].length; j++) {
+      testArr[i][j] = i * j;
+    }
+  }
+  for (let i = 0; i < testArr.length; i++) {
+    for (let j = 0; j < testArr[i].length; j++) {
+      testArr[i][j] = Math.sqrt(j * testArr[i][j]);
     }
   }
   if (Date.now() - startTime < 1000 * durationSeconds) {
-    setImmediate(() => benchmark(durationSeconds));
+    setTimeout(() => benchmark(durationSeconds), 5);
   }
 }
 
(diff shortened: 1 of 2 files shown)


testing/busybench.js
```diff
@@ -15,16 +15,18 @@
  */
 
 const startTime = Date.now();
+const test_arr = [];
```
aalexand: Naming style is off.

testing/busybench.js
```diff
@@ -15,16 +15,18 @@
...
- * Repeatedly fills a buffer, then calls itself with setImmediate.
+ * Fills several arrays, then calls itself with setImmediate.
  * It continues to do this until durationSeconds after the startTime.
  */
 function benchmark(durationSeconds) {
-  var buffer = new Buffer(1e4);
-  for (var k = 0; k < 1e4; k++) {
-    for (var j = 0; j < buffer.length; j++) {
-      buffer[j] = 0;
+  //_.test_buffer = new Buffer(1e8);
```
aalexand: Remove commented code.

testing/busybench.js
```diff
@@ -15,16 +15,18 @@
...
+  for (let i = 0; i < 10; i++) {
+    test_arr[i] = new Array(1e5);
```
aalexand: The allocation can be done once, doesn't need to be done on every benchmark() call.

testing/busybench.js
nolanmar511: done.

testing/busybench.js
nolanmar511: done.

testing/busybench.js
nolanmar511: Okay -- only make new arrays on the first call. 
I'm still storing the result of the computation in the array, to make sure it isn't optimized away.

testing/busybench.js
```diff
@@ -15,16 +15,23 @@
...
- * Repeatedly fills a buffer, then calls itself with setImmediate.
+ * Fills several arrays, then calls itself with setImmediate.
  * It continues to do this until durationSeconds after the startTime.
  */
 function benchmark(durationSeconds) {
-  var buffer = new Buffer(1e4);
-  for (var k = 0; k < 1e4; k++) {
-    for (var j = 0; j < buffer.length; j++) {
-      buffer[j] = 0;
+  if (callCount == 0) {
```
aalexand: Just make this a separate function like alloc() and call before benchmark()?

nolanmar511: When I separated this into a different function, that function didn't appear in the profile. I filed https://github.com/v8/sampling-heap-profiler/issues/16.

nolanmar511: My problems with using the alloc function were related to #160.

nolanmar511:
PTAL

Issue #116, opened 2018-01-19
Avoid require to load config file


ofrobots:
https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/a7f742d097dfab0cfdf11b0e5844dabb1cc09d65/ts/src/index.ts#L76

If the env. var. points to a .js or .node file, it will be executed. It would be better to read the file as text and then parse as json.

aalexand:
@nolanmar511 Did #161 fix this?

ofrobots:
This should be fixed. @nolanmar511 please double check and close the issue if so.

nolanmar511:
Yes, this is fixed by #161

PR #161, opened 2018-03-13
fix: start sampling heap profiler before any asynchronous calls

### LINKED ISSUE #160: Heap Profiles are missing Module stacks
Profiles taken with [sampling-heap-profiler](https://github.com/v8/sampling-heap-profiler/issues/16) have stacks which contain functions like `Module.runMain` and `Module._load`, like the following profile from ofrobots collected in https://github.com/v8/sampling-heap-profiler/issues/16:
image

cloud-profiler-nodejs does not collect with stacks with functions like `Module.runMain` or `Module._load`. 

This is likely because the [`start()`](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/86ef3ab605aea7a8d5d0750aef3224a6f7f8162d/ts/src/index.ts#L124) function for cloud-profiler-nodejs is asynchronous, so we can't collect samples on the first tick.

Right now, [`start()`](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/86ef3ab605aea7a8d5d0750aef3224a6f7f8162d/ts/src/index.ts#L124) must be asynchronous because [`initConfig()`](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/86ef3ab605aea7a8d5d0750aef3224a6f7f8162d/ts/src/index.ts#L55) is asynchronous, so it can retrieve instance and zone from metadata when those fields are not passed in.

nolanmar511:
fixes #160

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -17,6 +17,7 @@
 import {Logger, util} from '@google-cloud/common';
 import * as delay from 'delay';
 import * as extend from 'extend';
+import * as fs from 'fs';
 import * as gcpMetadata from 'gcp-metadata';
 import * as path from 'path';
 import {normalize} from 'path';
@@ -25,6 +26,7 @@ import * as semver from 'semver';
 
 import {Config, defaultConfig, ProfilerConfig} from './config';
 import {Profiler} from './profiler';
+import * as heapProfiler from './profilers/heap-profiler';
 
 const pjson = require('../../package.json');
 
@@ -45,12 +47,10 @@ function hasService(config: Config):
 
 /**
  * Sets unset values in the configuration to the value retrieved from
- * environment variables, metadata, or specified in defaultConfig.
+ * environment variables or specified in defaultConfig.
  * Throws error if value that must be set cannot be initialized.
- *
- * Exported for testing purposes.
  */
-export async function initConfig(config: Config): Promise<ProfilerConfig> {
+function initConfigLocal(config: Config): ProfilerConfig {
   config = util.normalizeArguments(null, config);
 
   const envConfig: Config = {
@@ -69,15 +69,38 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   }
 
   let envSetConfig: Config = {};
-  const val = process.env.GCLOUD_PROFILER_CONFIG;
-  if (val) {
-    envSetConfig = require(path.resolve(val)) as Config;
+  const configPath = process.env.GCLOUD_PROFILER_CONFIG;
+  if (configPath) {
+    let envSetConfigBuf;
+    try {
+      envSetConfigBuf = fs.readFileSync(configPath);
+    } catch (e) {
+      throw Error(`Could not read GCLOUD_PROFILER_CONFIG ${configPath}: ${e}`);
+    }
+    try {
+      envSetConfig = JSON.parse(envSetConfigBuf.toString());
+    } catch (e) {
+      throw Error(`Could not parse GCLOUD_PROFILER_CONFIG ${configPath}: ${e}`);
+    }
   }
 
   const mergedConfig =
       extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
-  if (!mergedConfig.zone || !mergedConfig.instance) {
+  if (!hasService(mergedConfig)) {
+    throw new Error('Service must be specified in the configuration.');
+  }
+
+  return mergedConfig;
+}
+
+/**
+ * Sets unset values in the configuration which can be retrieved from GCP
+ * metadata.
+ */
+async function initConfigMetadata(config: ProfilerConfig):
+    Promise<ProfilerConfig> {
+  if (!config.zone || !config.instance) {
     const [instance, zone] =
         await Promise
             .all([
@@ -88,22 +111,40 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
                     // ignore errors, which will occur when not on GCE.
                 }) ||
         [undefined, undefined];
-    if (!mergedConfig.zone && zone) {
-      mergedConfig.zone = zone.substring(zone.lastIndexOf('/') + 1);
+    if (!config.zone && zone) {
+      config.zone = zone.substring(zone.lastIndexOf('/') + 1);
     }
-    if (!mergedConfig.instance && instance) {
-      mergedConfig.instance = instance;
+    if (!config.instance && instance) {
+      config.instance = instance;
     }
   }
-
-  if (!hasService(mergedConfig)) {
-    throw new Error('Service must be specified in the configuration.');
-  }
-
-  return mergedConfig;
+  return config;
 }
 
-let profiler: Profiler|undefined = undefined;
+/**
+ * Initializes the config, and starts heap profiler if the heap profiler is
+ * needed. Returns a profiler if creation is successful. Otherwise, returns
+ * rejected promise.
+ */
+export async function createProfiler(config: Config): Promise<Profiler> {
+  if (!semver.satisfies(process.version, pjson.engines.node)) {
+    throw new Error(
+        `Could not start profiler: node version ${process.version}` +
+        ` does not satisfies "${pjson.engines.node}"`);
+  }
+
+  let profilerConfig: ProfilerConfig = initConfigLocal(config);
+
+  // Start the heap profiler if profiler config does not indicate heap profiling
+  // is disabled. This must be done before any asynchronous calls are made so
+  // all memory allocations made after start() is called can be captured.
+  if (!profilerConfig.disableHeap) {
+    heapProfiler.start(
+        profilerConfig.heapIntervalBytes, profilerConfig.heapMaxStackDepth);
+  }
+  profilerConfig = await initConfigMetadata(profilerConfig);
+  return new Profiler(profilerConfig);
+}
 
 /**
  * Starts the profiling agent and returns a promise.
@@ -120,21 +161,13 @@ let profiler: Profiler|undefined = undefined;
  *
  */
 export async function start(config: Config = {}): Promise<void> {
-  if (!semver.satisfies(process.version, pjson.engines.node)) {
-    logError(
-        `Could not start profiler: node version ${process.version}` +
-            ` does not satisfies "${pjson.engines.node}"`,
-        config);
-    return;
-  }
-  let normalizedConfig: ProfilerConfig;
+  let profiler: Profiler;
   try {
-    normalizedConfig = await initConfig(config);
+    profiler = await createProfiler(config);
   } catch (e) {
-    logError(`Could not start profiler: ${e}`, config);
+    logError(`${e}`, config);
     return;
   }
-  profiler = new Profiler(normalizedConfig);
   profiler.start();
 }
 
@@ -152,12 +185,17 @@ function logError(msg: string, config: Config) {
  * profiles.
  */
 export async function startLocal(config: Config = {}): Promise<void> {
-  const normalizedConfig = await initConfig(config);
-  const profiler = new Profiler(normalizedConfig);
+  let profiler: Profiler;
+  try {
+    profiler = await createProfiler(config);
+  } catch (e) {
+    logError(`${e}`, config);
+    return;
+  }
 
   // Set up periodic logging.
   const logger = new Logger({
-    level: Logger.DEFAULT_OPTIONS.levels[normalizedConfig.logLevel],
+    level: Logger.DEFAULT_OPTIONS.levels[profiler.config.logLevel],
     tag: pjson.name
   });
   let heapProfileCount = 0;
@@ -189,7 +227,7 @@ export async function startLocal(config: Config = {}): Promise<void> {
     heapProfileCount = 0;
     timeProfileCount = 0;
     prevLogTime = curTime;
-  }, normalizedConfig.localLogPeriodMillis);
+  }, profiler.config.localLogPeriodMillis);
 
   // Periodic profiling
   setInterval(async () => {
@@ -198,16 +236,16 @@ export async function startLocal(config: Config = {}): Promise<void> {
           {name: 'Heap-Profile' + new Date(), profileType: 'HEAP'});
       heapProfileCount++;
     }
-    await delay(normalizedConfig.localProfilingPeriodMillis / 2);
+    await delay(profiler.config.localProfilingPeriodMillis / 2);
     if (!config.disableTime) {
       const wall = await profiler.profile({
         name: 'Time-Profile' + new Date(),
         profileType: 'WALL',
-        duration: normalizedConfig.localTimeDurationMillis.toString() + 'ms'
+        duration: profiler.config.localTimeDurationMillis.toString() + 'ms'
       });
       timeProfileCount++;
     }
-  }, normalizedConfig.localProfilingPeriodMillis);
+  }, profiler.config.localProfilingPeriodMillis);
 }
 
 // If the module was --require'd from the command line, start the agent.
--- a/ts/src/profilers/heap-profiler.ts
+++ b/ts/src/profilers/heap-profiler.ts
@@ -15,41 +15,52 @@
  */
 
 import {perftools} from '../../../proto/profile';
+import {defaultConfig} from '../config';
+
 import {serializeHeapProfile} from './profile-serializer';
 
 const profiler = require('bindings')('sampling_heap_profiler');
 
-export class HeapProfiler {
-  private enabled = false;
+let enabled = false;
+let heapIntervalBytes = 0;
+let heapStackDepth = 0;
 
-  /**
-   * @param intervalBytes - average bytes between samples.
-   * @param stackDepth - upper limit on number of frames in stack for sample.
-   */
-  constructor(private intervalBytes: number, private stackDepth: number) {
-    this.enable();
+/*
+ * Collects a heap profile when heapProfiler is enabled. Otherwise throws
+ * an error.
+ */
+export function profile(): perftools.profiles.IProfile {
+  if (!enabled) {
+    throw new Error('Heap profiler is not enabled.');
   }
+  const startTimeNanos = Date.now() * 1000 * 1000;
+  const result = profiler.getAllocationProfile();
+  return serializeHeapProfile(result, startTimeNanos, heapIntervalBytes);
+}
 
-  /**
-   * Collects a heap profile when heapProfiler is enabled. Otherwise throws
-   * an error.
-   */
-  profile(): perftools.profiles.IProfile {
-    if (!this.enabled) {
-      throw new Error('Heap profiler is not enabled.');
-    }
-    const result = profiler.getAllocationProfile();
-    const startTimeNanos = Date.now() * 1000 * 1000;
-    return serializeHeapProfile(result, startTimeNanos, this.intervalBytes);
+/**
+ * Starts heap profiling. If heap profiling has already been started with
+ * the same parameters, this is a noop. If heap profiler has already been
+ * started with different parameters, this throws an error.
+ *
+ * @param intervalBytes - average number of bytes between samples.
+ * @param stackDepth - maximum stack depth for samples collected.
+ */
+export function start(intervalBytes: number, stackDepth: number) {
+  if (enabled) {
+    throw new Error(`Heap profiler is already started  with intervalBytes ${
+        heapIntervalBytes} and stackDepth ${stackDepth}`);
   }
+  heapIntervalBytes = intervalBytes;
+  heapStackDepth = stackDepth;
+  profiler.startSamplingHeapProfiler(heapIntervalBytes, heapStackDepth);
+  enabled = true;
+}
 
-  enable() {
-    profiler.startSamplingHeapProfiler(this.intervalBytes, this.stackDepth);
-    this.enabled = true;
-  }
-
-  disable() {
-    this.enabled = false;
+// Stops heap profiling. If heap profiling has not been started, does nothing.
+export function stop() {
+  if (enabled) {
+    enabled = false;
     profiler.stopSamplingHeapProfiler();
   }
 }
--- a/ts/test/test-heap-profiler.ts
+++ b/ts/test/test-heap-profiler.ts
@@ -18,7 +18,7 @@ import * as delay from 'delay';
 import * as sinon from 'sinon';
 
 import {perftools} from '../../proto/profile';
-import {HeapProfiler} from '../src/profilers/heap-profiler';
+import * as heapProfiler from '../src/profilers/heap-profiler';
 
 import {heapProfile, v8HeapProfile} from './profiles-for-tests';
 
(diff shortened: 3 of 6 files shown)


ts/src/index.ts
```diff
@@ -29,6 +29,7 @@ import {Profiler} from './profiler';
 
 const common: Common = require('@google-cloud/common');
 const pjson = require('../../package.json');
+const heapProfiler = require('bindings')('sampling_heap_profiler');
```
aalexand: It doesn't seem right to call directly into the binding here. The profilers/heap-profiler should be the only level of abstraction on top of that, so call that.

BTW, heap-profiler still calls enable() from its constructor? Doesn't seem right if so - not sure which effects calling that twice has.

ts/src/index.ts
```diff
@@ -129,13 +130,28 @@ export async function start(config: Config = {}): Promise<void> {
...
+    heapProfiler.startSamplingHeapProfiler(
+        config.heapIntervalBytes || defaultConfig.heapIntervalBytes,
+        config.heapMaxStackDepth || defaultConfig.heapIntervalBytes);
+  }
+
   let normalizedConfig: ProfilerConfig;
   try {
     normalizedConfig = await initConfig(config);
   } catch (e) {
+    heapProfiler.stopSamplingHeapProfiler();
```
aalexand: Is it right that it's unconditional?

ts/src/index.ts
```diff
@@ -152,7 +168,21 @@ function logError(msg: string, config: Config) {
  * profiles.
  */
 export async function startLocal(config: Config = {}): Promise<void> {
+  // start the v8 heap profiler if heap profiling is enabled. This must be done
+  // before any asynchronous calls are made so samples from the first tick can
+  // be captured.
+  if (!(config.disableHeap === false)) {
```
aalexand: Can it be just `if (!!config.disableHeap)`?

ts/src/index.ts
```diff
@@ -129,13 +130,28 @@ export async function start(config: Config = {}): Promise<void> {
...
   try {
     normalizedConfig = await initConfig(config);
   } catch (e) {
+    heapProfiler.stopSamplingHeapProfiler();
     logError(`Could not start profiler: ${e}`, config);
     return;
   }
+
+  if (normalizedConfig.disableHeap) {
+    heapProfiler.stopSamplingHeapProfiler();
```
aalexand: Why is it stopped here? If this is sequence like "enable - disable - enable in the constructor" then it's pretty convoluted, any chance to simplify?

ts/src/index.ts
aalexand: I think you want it in finally{}, not in catch{}. Then you can have it just once.

ts/src/index.ts
nolanmar511: I've switched to creating a HeapProfiler, which is passed in to the profiler.

ts/src/index.ts
nolanmar511: This is now conditional.

ts/src/index.ts

ts/src/index.ts
nolanmar511: I think I've simplified this.

ts/src/index.ts
```diff
@@ -129,14 +130,36 @@ export async function start(config: Config = {}): Promise<void> {
         config);
     return;
   }
+
+  // start the heap profiler if profiler config does not indicate heap profiling
+  // is disabled. This must be done before any asynchronous calls are made so
+  // samples from the first tick can be captured.
+  let heapProfiler: HeapProfiler|undefined;
+  if (!!config.disableHeap) {
```
ofrobots: Should this be a single negation?

ts/src/index.ts
```diff
@@ -152,9 +175,22 @@ function logError(msg: string, config: Config) {
  * profiles.
  */
 export async function startLocal(config: Config = {}): Promise<void> {
+  // start the v8 heap profiler if heap profiling is enabled. This must be done
+  // before any asynchronous calls are made so samples from the first tick can
+  // be captured.
+  let heapProfiler: HeapProfiler|undefined;
+  if (!!config.disableHeap) {
```
ofrobots: ditto

nolanmar511:
Sorry -- pushed to this branch by mistake at some point. 
The branch is now correct. PTAL

ts/src/index.ts
```diff
@@ -129,14 +130,36 @@ export async function start(config: Config = {}): Promise<void> {
         config);
     return;
   }
+
+  // start the heap profiler if profiler config does not indicate heap profiling
```
aalexand: Nit: start -> Start

ts/src/index.ts
```diff
@@ -129,14 +130,36 @@ export async function start(config: Config = {}): Promise<void> {
...
+  // is disabled. This must be done before any asynchronous calls are made so
+  // samples from the first tick can be captured.
+  let heapProfiler: HeapProfiler|undefined;
+  if (!config.disableHeap) {
+    heapProfiler = new HeapProfiler(
```
aalexand: Creating the heap profiler seems overkill. Can you have enable() / disable() methods as static on the heap profiler class (or just as functions) instead and call those?

aalexand: Also make enable() / disable() set and check the global boolean flag for whether the heap profiler is enabled. Then you can call disable() without a check around it.

ts/src/index.ts

ts/src/index.ts
nolanmar511: I removed the heap profiler class. I think doing this addresses this comments, an also makes sense because (unlike the time profiler) the heap profiler can only be collecting a profile with one set of parameters.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -129,13 +130,34 @@ export async function start(config: Config = {}): Promise<void> {
         config);
     return;
   }
+
+  // Start the heap profiler if profiler config does not indicate heap profiling
+  // is disabled. This must be done before any asynchronous calls are made so
+  // samples from the first tick can be captured.
+  if (!config.disableHeap) {
+    heapProfiler.set(
```
aalexand: Requesting some changes:
* Remove set() method, pass parameters to enable().
* Rename enable() -> start(), disable() -> stop().
* Make stop() no op if start() wasn't called. Document it so. Remove isEnabled() method and just call stop() when needed.

This all should simplify code.

ts/src/index.ts
```diff
@@ -129,13 +130,34 @@ export async function start(config: Config = {}): Promise<void> {
         config);
     return;
   }
+
+  // Start the heap profiler if profiler config does not indicate heap profiling
+  // is disabled. This must be done before any asynchronous calls are made so
+  // samples from the first tick can be captured.
```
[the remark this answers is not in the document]
aalexand: I think the requirement about the first tick is too strict. What I'd expect is that the heap profiling starts as soon as the require / start is called in the source and everything that is executed after / below that is heap-profiled.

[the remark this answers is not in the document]
nolanmar511: I clarified this comment. All code/allocations after .start() is called should be profiled.

ts/src/index.ts
```diff
@@ -129,13 +130,31 @@ export async function start(config: Config = {}): Promise<void> {
...
   let normalizedConfig: ProfilerConfig;
   try {
     normalizedConfig = await initConfig(config);
   } catch (e) {
+    heapProfiler.stop();
     logError(`Could not start profiler: ${e}`, config);
     return;
   }
+
+  // stop heap profiler if, after initialization, the config indicates that
```
aalexand: Nit: stop -> Stop

ts/src/index.ts
```diff
@@ -152,9 +171,24 @@ function logError(msg: string, config: Config) {
  * profiles.
  */
 export async function startLocal(config: Config = {}): Promise<void> {
+  // Start the v8 heap profiler if heap profiling is enabled. This must be done
```
aalexand: It feels that code from here up to "profiler = new Profiler(normalizedConfig);" could be shared between start() and startLocal()?

ts/src/profilers/heap-profiler.ts
```diff
@@ -15,41 +15,54 @@
...
-   */
-  profile(): perftools.profiles.IProfile {
-    if (!this.enabled) {
-      throw new Error('Heap profiler is not enabled.');
-    }
-    const result = profiler.getAllocationProfile();
-    const startTimeNanos = Date.now() * 1000 * 1000;
-    return serializeHeapProfile(result, startTimeNanos, this.intervalBytes);
+/**
+ * Starts heap profiling. If heap profiling has already been started with
```
aalexand: Optional: I think you could as well return an error if an attempt to start already started profiler with different parameters is made as I don't think it should ever be done with the current agent.

nolanmar511: This is possible with the code as written.
If the user specifies heapIntervalBytes or heapMaxStackDepth in the config pointed to by GCLOUD_PROFILER_CONFIG, then [this](https://github.com/nolanmar511/cloud-profiler-nodejs/blob/193761afd9d42abd41e0b06dc2a175672657390a/ts/src/profiler.ts#L261-L262) line in startProfiler would start the profiler with different parameters... and the profiles could be off.

I can work around this, but only if we decide to never fix #116, because loading the config with require() makes it possible to load the config synchronously.

aalexand:
Is this ready to be re-reviewed? Also, the coverage goes down, can tests be improved to make that go up or stay the same?

nolanmar511:
I think I have one question which needs to be resolved before I make the changes to re-review this:

If the user specifies heapIntervalBytes or heapMaxStackDepth in the config pointed to by GCLOUD_PROFILER_CONFIG, then this line in startProfiler would start the profiler with different parameters... and the profiles could be off, or may end up missing allocations (I need to still investigate to determine the impact).

I can work around this, but only if we decide to not to fix #116, because loading the config with require() makes it possible to load the config synchronously.

I think loading the config with require is reasonable since [trace](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/c8c5bfc6168649cba65ca32cf53a524d38dac521/src/index.ts#L83-L84) still does this, and I'm guessing that, since trace needs to be started first, having an asynchronous piece of code when initializing the config would not work for trace.

aalexand:
@nolanmar511 When we get rid of using `require`, could we use [readFileSync](https://nodejs.org/api/fs.html#fs_fs_readfilesync_path_options) to read the config file contents? It's probably what `require` does underneath to be synchronous anyway.

nolanmar511:
PTAL
I think planning to use readFileSync if a user specifies a config file is reasonable.

ts/src/index.ts
```diff
@@ -90,23 +101,37 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
+  return config;
 }
 
 let profiler: Profiler|undefined = undefined;
 
+/**
+ * Initializes the config, and starts heap profiler if the heap profiler is
+ * needed.
+ */
+export async function startHelper(config: Config): Promise<ProfilerConfig> {
```
aalexand: Would be good to name it more specifically than `startHelper`. Maybe `prepareStart` or `preStart` or even `initConfigAndMaybeStartHeapProfiler`?

ts/src/index.ts
```diff
@@ -90,23 +101,37 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
+  let profilerConfig = initConfig(config);
+
+  // Start the heap profiler if profiler config does not indicate heap profiling
+  // is disabled. This must be done before any asynchronous calls are made so
+  // all memory allocations made after start() is called can be captured.
+  if (!config.disableHeap) {
```
aalexand: Is this intentional that this checks config and not profilerConfig? If so, why it's not before the initConfig() call?

ts/src/profilers/heap-profiler.ts
aalexand: Should this return an error now if start() is called after the profiler is already started?

nolanmar511: I think that makes sense.
Done.

ts/src/index.ts
nolanmar511: I went with `initConfigAndMaybeStartHeapProfiler`.

ts/src/index.ts
nolanmar511: Fixed.

aalexand:
@nolanmar511 Ready to PTAL?

ts/src/index.ts
aalexand: It doesn't look like the function is renamed? I'm still seeing startHelper.

ts/src/index.ts
aalexand: Not fixed?

ts/src/profilers/heap-profiler.ts
aalexand: Not done?

ts/src/index.ts
```diff
@@ -129,13 +155,21 @@ export async function start(config: Config = {}): Promise<void> {
...
-    normalizedConfig = await initConfig(config);
+    normalizedConfig = await initConfigAndMaybeStartHeapProfiler(config);
   } catch (e) {
     logError(`Could not start profiler: ${e}`, config);
     return;
   }
+
+  // Stop heap profiler if, after initialization, the config indicates that
+  // the heap profiler is disabled.
+  if (normalizedConfig.disableHeap) {
```
ofrobots: I don't think this case is possible any more, or is it? `disableHeap` can be synchronously configured only. `initConfigMetadata` cannot change the value of `disableHeap`.

Either remove, or change this to an assertion?

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -129,13 +155,15 @@ export async function start(config: Config = {}): Promise<void> {
         config);
     return;
   }
+
   let normalizedConfig: ProfilerConfig;
```
aalexand: Looking more into it now, it feels you could move all of
```
let normalizedConfig: ProfilerConfig;
...
profiler = new Profiler(normalizedConfig);
```
into `initConfigAndMaybeStartHeapProfiler` naming it as createProfile() which returns the created profiler or null and call it as "profiler = createProfiler(); if (!profiler) return;" (the logging would be inside of createProfiler).

ts/src/index.ts
```diff
@@ -129,13 +155,15 @@ export async function start(config: Config = {}): Promise<void> {
...
   try {
-    normalizedConfig = await initConfig(config);
+    normalizedConfig = await initConfigAndMaybeStartHeapProfiler(config);
   } catch (e) {
     logError(`Could not start profiler: ${e}`, config);
```
aalexand: start -> initialize?

ts/src/index.ts

ts/src/index.ts

aalexand:
@nolanmar511 Did you push the changes? I don't see the "Done" things in the source.

nolanmar511:
Haven't pushed the changes. Will add PTAL when I have.

nolanmar511:
PTAL

aalexand:
@ofrobots Could you LGTM as well?

ts/src/index.ts
```diff
@@ -79,7 +78,20 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   const mergedConfig =
       extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
-  if (!mergedConfig.zone || !mergedConfig.instance) {
+  if (!hasService(mergedConfig)) {
```
ofrobots: Q: does this change make this a semver major change? That is, previously working code may no longer work if it relied on the service context being discovered from the metadata service?

ts/src/index.ts
```diff
@@ -90,22 +102,51 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
+ * undefined.
+ */
+export async function createProfiler(config: Config):
+    Promise<Profiler|undefined> {
+  if (!semver.satisfies(process.version, pjson.engines.node)) {
+    logError(
+        `Could not start profiler: node version ${process.version}` +
+            ` does not satisfies "${pjson.engines.node}"`,
+        config);
+    return undefined;
```
ofrobots: nit: it would be better to throw here. This would get rid of the need to declare the return type to be the ungainly `Profiler|undefined`.

ts/src/index.ts
```diff
@@ -90,22 +102,51 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
   }
 
-  return mergedConfig;
-}
+  let profilerConfig: ProfilerConfig;
+  try {
+    profilerConfig = initConfigLocal(config);
+  } catch (e) {
+    logError(`Could not start profiler: ${e}`, config);
+    return undefined;
```
ofrobots: Likewise. Omit the catch block here, and let the throw reach the caller.

ts/src/index.ts
```diff
@@ -122,21 +163,10 @@ let profiler: Profiler|undefined = undefined;
...
-        config);
-    return;
-  }
-  let normalizedConfig: ProfilerConfig;
-  try {
-    normalizedConfig = await initConfig(config);
-  } catch (e) {
-    logError(`Could not start profiler: ${e}`, config);
+  const profiler: Profiler|undefined = await createProfiler(config);
+  if (profiler === undefined) {
```
ofrobots: this test can be removed if you throw above. Omit the type for the local variable `profiler`.

ts/src/index.ts
```diff
@@ -152,8 +182,10 @@ function logError(msg: string, config: Config) {
  * profiles.
  */
 export async function startLocal(config: Config = {}): Promise<void> {
-  const normalizedConfig = await initConfig(config);
-  profiler = new Profiler(normalizedConfig);
+  const profiler: Profiler|undefined = await createProfiler(config);
+  if (profiler === undefined) {
```
ofrobots: Likewise.

ts/test/test-heap-profiler.ts
```diff
@@ -15,53 +15,116 @@
  */
 
 import * as delay from 'delay';
+import {describe} from 'mocha';
```
ofrobots: Why was this needed? We were using `describe` before just fine. Note that we are not importing any of the other globals added by `mocha` (e.g. `beforeEach`).

ts/src/index.ts
nolanmar511: We've never checked metadata for the service, so this would not be semver major.

ts/src/index.ts
```diff
@@ -116,25 +123,21 @@ const profiler: Profiler|undefined = undefined;
...
+        ` does not satisfies "${pjson.engines.node}"`);
   }
 
   let profilerConfig: ProfilerConfig;
   try {
     profilerConfig = initConfigLocal(config);
   } catch (e) {
-    logError(`Could not start profiler: ${e}`, config);
-    return undefined;
+    throw new Error(`Could not start profiler: ${e}`);
```
ofrobots: It would be better to throw the original error. The additional message doesn't really add value. That is, I think you can omit the try/catch.

ts/src/index.ts
```diff
@@ -72,6 +73,12 @@ function initConfigLocal(config: Config): ProfilerConfig {
   let envSetConfig: Config = {};
   const val = process.env.GCLOUD_PROFILER_CONFIG;
   if (val) {
+    try {
+      const envSetConfigBuf = fs.readFileSync(val);
+      envSetConfig = JSON.parse(envSetConfigBuf.toString());
```
ofrobots: It is good to keep your `try` block focussed. I think you are intending only to catch the the `JSON.parse()` error here. Reduce the `try` block to just that.

ts/src/index.ts
```diff
@@ -116,25 +123,21 @@ const profiler: Profiler|undefined = undefined;
 
 /**
  * Initializes the config, and starts heap profiler if the heap profiler is
- * needed. Returns a profiler if creation is successful. Otherwise, returns
- * undefined.
+ * needed. Returns a profiler if creation is successful. Otherwise, throws an
+ * error.
```
ofrobots: The return value is actually a promise, so this function never throws actually. It will always be a promise – a rejected promise in the case of a throw.

ts/src/index.ts
nolanmar511: Done

ts/src/index.ts
nolanmar511: done

ts/src/index.ts
nolanmar511: done

ts/src/index.ts
nolanmar511: done

ts/test/test-heap-profiler.ts
nolanmar511: removed -- this isn't needed

ts/src/index.ts

ts/src/index.ts
nolanmar511: Removed try/catch block.
I think it may make sense to just let the error be thrown.

ts/src/index.ts
nolanmar511: changed comment.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -73,13 +73,28 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   let envSetConfig: Config = {};
   const val = process.env.GCLOUD_PROFILER_CONFIG;
   if (val) {
+    const envSetConfigBuf = fs.readFileSync(val);
```
aalexand: This will throw an exception if val points to a non-existing / non-readable file. Should this be handled?

ts/src/index.ts
```diff
@@ -73,13 +73,28 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
   let envSetConfig: Config = {};
   const val = process.env.GCLOUD_PROFILER_CONFIG;
```
aalexand: Rename val to "path"?

ts/src/profilers/heap-profiler.ts
```diff
@@ -15,41 +15,52 @@
...
-
-  /**
-   * @param intervalBytes - average bytes between samples.
-   * @param stackDepth - upper limit on number of frames in stack for sample.
-   */
-  constructor(private intervalBytes: number, private stackDepth: number) {
-    this.enable();
-  }
+let enabled = false;
+let heapIntervalBytes = defaultConfig.heapIntervalBytes;
```
aalexand: I think you could initialize heapIntervalBytes and heapStackDepth to zeros here. Any reason to depend on config?

ts/src/profilers/heap-profiler.ts
```diff
@@ -15,41 +15,52 @@
...
+/*
+ * Collects a heap profile when heapProfiler is enabled. Otherwise throws
+ * an error.
+ */
+export function profile(): perftools.profiles.IProfile {
+  if (!enabled) {
+    throw new Error('Heap profiler is not enabled.');
   }
+  const result = profiler.getAllocationProfile();
+  const startTimeNanos = Date.now() * 1000 * 1000;
```
aalexand: It would be slightly better to record the start time before calling getAllocationProfile maybe?

aalexand:
@nolanmar511 Could you take a pass over the remaining minor comments?

ts/src/profilers/heap-profiler.ts
nolanmar511: done.

ts/src/profilers/heap-profiler.ts
nolanmar511: done.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -90,22 +113,41 @@ export async function initConfig(config: Config): Promise<ProfilerConfig> {
...
+export async function createProfiler(config: Config): Promise<Profiler> {
+  if (!semver.satisfies(process.version, pjson.engines.node)) {
+    throw new Error(
+        `Could not start profiler: node version ${process.version}` +
+        ` does not satisfies "${pjson.engines.node}"`);
   }
 
-  return mergedConfig;
-}
+  let profilerConfig: ProfilerConfig;
```
aalexand: This and next line could be single `let profilerConfig = initConfigLocal(config);` line?

ts/test/test-init-config.ts
```diff
@@ -137,12 +170,20 @@ describe('initConfig', () => {
          disableTime: false,
          projectId: 'fake-projectId',
        };
-       const initializedConfig = await initConfig(config);
+       const profiler: Profiler|undefined = await createProfiler(config);
+       if (profiler === undefined) {
+         assert.ok(true, 'profiler should be created successfully.');
```
ofrobots: The only case where `profiler` would be undefined is when `createProfiler` throws. What is this test trying to achieve?

This comment applies to all the places where this test is being done. Ideally, the type for the variable `profiler` should be `Profiler` rather than `Profiler|undefined`, but given that's the declared type of the function anyway, you should probably omit the type in the variable declaration here.

ts/test/test-init-config.ts
```diff
@@ -70,7 +79,15 @@ describe('initConfig', () => {
...
-    const initializedConfig = await initConfig(config);
+    const profiler: Profiler|undefined = await createProfiler(config);
+    if (profiler === undefined) {
+      assert.ok(true, 'profiler should be created successfully.');
+      return;
+    }
+    const initializedConfig = profiler.config;
+    // remove interceptors_ field which is added to profiler.config.
+    // tslint:disable-next-line: no-any
+    delete (initializedConfig as any).interceptors_;
```
ofrobots: It might better to call `common.util.normalizeConfig` on `internalConfigParams` above and avoid this complexity here in each of the tests.

nolanmar511: Done.

ts/test/test-init-config.ts
nolanmar511: Done.

nolanmar511:
PTAL

ts/test/test-init-config.ts
```diff
@@ -70,8 +82,10 @@ describe('initConfig', () => {
       zone: 'zone',
       projectId: 'fake-projectId'
     };
-    const initializedConfig = await initConfig(config);
-    assert.deepEqual(initializedConfig, extend(config, internalConfigParams));
+    const profiler: Profiler = await createProfiler(config);
+    const expConfig = await common.util.normalizeArguments(
```
ofrobots: suggestion: would it be possible to move `normalizeArugments` call to an outer scope, and use the following here:
```ts
const expConfig = extend({},DEFAULT_CONFIG, config);
```

where `DEFAULT_CONFIG` could be:
```ts
const DEFAULT_CONFIG = await common.util.normalizeArguments(null, internalConfigParams);
Object.freeze(DEFAULT_CONFIG);
```

nolanmar511: It doesn't seem to be possible. 
I don't know why, but when I do this, I get a lot of:
```
     TypeError: Attempted to wrap instance which is already wrapped
```

nolanmar511: [patch.patch.zip](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/files/1962425/patch.patch.zip)

ofrobots: I think the reason it doesn't work is because mocha probably doesn't support `describe` blocks to be async.

I would suggest a patch along the lines of:

```diff
diff --git a/ts/test/test-init-config.ts b/ts/test/test-init-config.ts
index 4082ad588..27af2c09d 100644
--- a/ts/test/test-init-config.ts
+++ b/ts/test/test-init-config.ts
@@ -33,7 +33,23 @@ describe('createProfiler', async () => {
   let metadataStub: sinon.SinonStub|undefined;
   let startStub: sinon.SinonStub;
 
-  before(() => {
+  const internalConfigParams = {
+    timeIntervalMicros: 1000,
+    heapIntervalBytes: 512 * 1024,
+    heapMaxStackDepth: 64,
+    initialBackoffMillis: 1000 * 60,
+    backoffCapMillis: 60 * 60 * 1000,
+    backoffMultiplier: 1.3,
+    serverBackoffCapMillis: 2147483647,
+    localProfilingPeriodMillis: 1000,
+    localTimeDurationMillis: 1000,
+    localLogPeriodMillis: 10000,
+    baseApiUrl: 'https://cloudprofiler.googleapis.com/v2',
+  };
+  let defaultConfig : typeof internalConfigParams;
+
+  before(async () => {
+    defaultConfig = await common.util.normalizeArguments(null, extend({}, internalConfigParams));
     startStub = sinon.stub(v8HeapProfiler, 'startSamplingHeapProfiler');
     savedEnv = process.env;
   });
@@ -55,19 +71,8 @@ describe('createProfiler', async () => {
     startStub.restore();
   });
 
-  const internalConfigParams = {
-    timeIntervalMicros: 1000,
-    heapIntervalBytes: 512 * 1024,
-    heapMaxStackDepth: 64,
-    initialBackoffMillis: 1000 * 60,
-    backoffCapMillis: 60 * 60 * 1000,
-    backoffMultiplier: 1.3,
-    serverBackoffCapMillis: 2147483647,
-    localProfilingPeriodMillis: 1000,
-    localTimeDurationMillis: 1000,
-    localLogPeriodMillis: 10000,
-    baseApiUrl: 'https://cloudprofiler.googleapis.com/v2',
-  };
+
+
 
   it('should not modify specified fields when not on GCE', async () => {
     metadataStub = sinon.stub(gcpMetadata, 'instance')
@@ -83,8 +88,9 @@ describe('createProfiler', async () => {
       projectId: 'fake-projectId'
     };
     const profiler: Profiler = await createProfiler(config);
-    const expConfig = await common.util.normalizeArguments(
-        null, extend({}, internalConfigParams, config));
+    // const expConfig = await common.util.normalizeArguments(
+    //     null, extend({}, internalConfigParams, config));
+    const expConfig = Object.assign({}, defaultConfig, config);
     assert.deepEqual(profiler.config, expConfig);
   });
 

```

nolanmar511: Done.

nolanmar511:
PTAL

PR #162, opened 2018-03-15
fix: update integration test to work as presubmit check

nolanmar511:
Kokoro indicates the PR being checked by PR number.
This change modified the integration test so PR number can be used to check out source code during the integration test.


testing/integration_test.go
```diff
@@ -103,19 +105,31 @@ type nodeGCETestCase struct {
 
 func (tc *nodeGCETestCase) initializeStartUpScript(template *template.Template) error {
 	var buf bytes.Buffer
+
+	var checkoutBranchCmd string
+	if *pr != 0 {
+		checkoutBranchCmd = fmt.Sprintf(
+			`git fetch origin pull/%d/head:pull_branch
```
aalexand: This became a strange mix of using templates and formatting strings manually to my taste. I'd suggest passing both pr and branch to the template and use template conditions to format the commands the needed way. See "if" pipeline in https://golang.org/pkg/text/template/#hdr-Actions.

testing/presubmit.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
aalexand: You could probably share some pieces of this file with continuous.sh. Or use the same script and condition slightly depending on whether KOKORO_GITHUB_PULL_REQUEST_NUMBER is set?

testing/integration_test.go
nolanmar511: Done.

testing/presubmit.sh
nolanmar511: Done.

testing/integration_test.go
```diff
@@ -65,9 +66,15 @@ npm -v
 node -v
 
 # Install agent
-git clone https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs.git
+git clone {{.Repo}}
 cd cloud-profiler-nodejs
-git pull {{.Repo}} {{.Branch}}
+{{if .PR}}
```
aalexand: I think you could do simpler:
```
git fetch origin pull/{{if .PR}}{{.PR}}{{else}}{{.Branch}}{{end}}/head:pull_branch
git checkout pull_branch
```

nolanmar511: done.

nolanmar511:
PTAL

PR #176, opened 2018-04-06
Add googletest as subtree

nolanmar511:


ofrobots:
This looks like a rubberstamp to me, but @aalexand should approve if this is the way to vs. submodule.

PR #178, opened 2018-04-13
feat: make profile duration and time between profiles configurable with startLocal()

nolanmar511:


ts/src/config.ts
```diff
@@ -106,6 +106,14 @@ export interface Config extends AuthenticationConfig {
   // Allows user to specify API URL other than
   // https://cloudprofiler.googleapis.com/v2.
   baseApiUrl?: string;
+
+  // Time to pause between profile collection.
+  // For testing with startLocal() only.
+  profileCollectionPauseMillis?: number;
```
aalexand: Let's prefix with "local", like "localPauseMillis".

ts/src/config.ts
```diff
@@ -106,6 +106,14 @@ export interface Config extends AuthenticationConfig {
...
+
+  // Duration of time profiles collected when using startLocal().
+  // For testing with startLocal() only.
+  timeDurationMillis?: number;
```
aalexand: Similar - something like "localTimeDurationMillis". "local time" reads a bit weird, like it's time in local time zone, but oh well. Suggest a better name if you can think of something.

ts/src/index.ts
```diff
@@ -155,19 +156,50 @@ export async function startLocal(config: Config = {}): Promise<void> {
...
+  setInterval(() => {
+    const curTime = Date.now();
+    const {rss, heapTotal, heapUsed} = process.memoryUsage();
+    logger.debug(
+        new Date().toISOString(), 'rss', numeral(rss).format('0.0 ib'),
+        'heapTotal', numeral(heapTotal).format('0.0 ib'), 'heapUsed',
+        numeral(heapUsed).format('0.0 ib'));
+    logger.debug(
+        new Date().toISOString(), heapProfileCount,
+        'heap profiles collectin in', curTime - prevLogTime, 'ms');
```
aalexand: "collectin" - typo?

ts/src/index.ts
```diff
@@ -155,19 +156,50 @@ export async function startLocal(config: Config = {}): Promise<void> {
...
+    logger.debug(
+        new Date().toISOString(), timeProfileCount,
+        'time profiles collectin in', curTime - prevLogTime, 'ms');
```
aalexand: Same typo?

package.json
```diff
@@ -32,6 +32,7 @@
     "extend": "^3.0.1",
     "gcp-metadata": "^0.6.1",
     "nan": "^2.8.0",
+    "numeral": "^2.0.6",
```
aalexand: I'd avoid dependencies that are not crucial, to minimize the profiler footprint.

ts/src/config.ts
nolanmar511: done.

ts/src/index.ts
nolanmar511: done.

ts/src/index.ts
nolanmar511: done.

package.json
nolanmar511: done.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -155,19 +155,49 @@ export async function startLocal(config: Config = {}): Promise<void> {
...
+  });
+  let heapProfileCount = 0;
+  let timeProfileCount = 0;
+  let prevLogTime = Date.now();
+
+  setInterval(() => {
+    const curTime = Date.now();
+    const {rss, heapTotal, heapUsed} = process.memoryUsage();
+    logger.debug(
+        new Date().toISOString(), 'rss', rss, 'B', 'heapTotal', heapTotal, 'B',
```
aalexand: Using KiB unit would probably make output more readable.

ts/src/index.ts
```diff
@@ -155,19 +155,49 @@ export async function startLocal(config: Config = {}): Promise<void> {
...
+
+    heapProfileCount = 0;
+    timeProfileCount = 0;
+    prevLogTime = curTime;
+  }, 10000);
+
+
   while (true) {
     if (!config.disableHeap) {
       const heap = await profiler.profile(
```
ofrobots: You're not changing this here, but it seems like currently we collect a Heap and then a Time profile serially but in a dependent fashion. Is this intentional? The duration between two successive heap profiles would be `localPauseMillis + ` localTimeDurationMillis`. The time profiler will never be able to profile the heap profiler. This might not be optimal.

Do we want two independent intervals for each of these profilers?

aalexand: Running the two profilers sequentially is intentional - the startLocal thing is intended to emulate what the API-curated mode does, just in a much more frequent succession. Profiling heap profiler with time profiler would not be what the main flow of the profiler does, and is not too useful anyway since most of the interesting stuff happens in the heap profiler in native code.

It would make sense though to wait for the specified duration after each profile though. Maybe wait for (pause + timeDuration) after heap profile collection so that the start points of each profile are equally distanced.

ts/src/index.ts
```diff
@@ -168,14 +168,13 @@ export async function startLocal(config: Config = {}): Promise<void> {
...
-        'heapUsed', heapUsed, 'B');
-    logger.debug(
-        new Date().toISOString(), heapProfileCount,
-        'heap profiles collected in', curTime - prevLogTime, 'ms');
-    logger.debug(
-        new Date().toISOString(), timeProfileCount,
-        'time profiles collected in', curTime - prevLogTime, 'ms');
+      new Date().toISOString(),
+      'rss', rss/(1024 * 1024), 'MiB,',
+      'heap total',  heapTotal/(1024 * 1024), 'MiB,',
```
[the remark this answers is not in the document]
nolanmar511: Fixed -- ran formatter.

ts/src/index.ts
nolanmar511: I switched to using setInterval rather than a while loop, which will, I think, provide the best assurances that profiles are collected at a specific rate.

ts/src/index.ts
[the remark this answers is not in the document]
nolanmar511: Now show number to 3 decimal places.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -153,22 +153,60 @@ function logError(msg: string, config: Config) {
...
+    tag: pjson.name
+  });
+  let heapProfileCount = 0;
+  let timeProfileCount = 0;
+  let prevLogTime = Date.now();
+
+  setInterval(() => {
+    const curTime = Date.now();
+    const {rss, heapTotal, heapUsed} = process.memoryUsage();
+    logger.debug(
```
aalexand: What the example output looks like? It appears there are no any separators between metric name and value. Want to make sure it looks readable.

ts/src/index.ts
```diff
@@ -153,22 +153,60 @@ function logError(msg: string, config: Config) {
...
+        'profiles/s',
+        'time profile collection rate',
+        (timeProfileCount * 1000 / (curTime - prevLogTime)).toFixed(3),
+        'profiles/s',
+    );
+
+    heapProfileCount = 0;
+    timeProfileCount = 0;
+    prevLogTime = curTime;
+  }, 10000);
```
aalexand: Make this interval configurable?

ts/src/index.ts
```diff
@@ -153,22 +153,60 @@ function logError(msg: string, config: Config) {
...
+  // Periodic profiling
+  setInterval(async () => {
     if (!config.disableHeap) {
       const heap = await profiler.profile(
           {name: 'HEAP-Profile' + new Date(), profileType: 'HEAP'});
+      heapProfileCount++;
     }
     if (!config.disableTime) {
       const wall = await profiler.profile({
         name: 'Time-Profile' + new Date(),
```
aalexand: Nit: is there a reason for different capitalization in "HEAP-Profile" vs. "Time-Profile"?

ts/src/index.ts
```diff
@@ -153,22 +153,60 @@ function logError(msg: string, config: Config) {
...
         profileType: 'WALL',
-        duration: '10s'
+        duration: normalizedConfig.localTimeDurationMillis.toString() + 'ms'
       });
+      timeProfileCount++;
     }
-    await delay(1000);
-  }
+  }, normalizedConfig.localProfilingPeriodMillis);
```
aalexand: The profile timing is still arranged in a way that the time profile is collected immediately after taking a heap profile. Having profiles equally distanced would be better, but I don't really think this fish is important given the size of the fishes we are trying to catch.

ts/src/index.ts
nolanmar511: Sample output:

```
DEBUG:@google-cloud/profiler: 2018-04-18T16:47:00.579Z rss 97.598 MiB, heap total 62.840 MiB, heap used 25.339 MiB, heap profile collection rate 338.900 profiles/s time profile collection rate 0.000 profiles/s
DEBUG:@google-cloud/profiler: 2018-04-18T16:47:10.580Z rss 98.879 MiB, heap total 63.840 MiB, heap used 22.269 MiB, heap profile collection rate 356.800 profiles/s time profile collection rate 0.000 profiles/s
DEBUG:@google-cloud/profiler: 2018-04-18T16:47:20.581Z rss 98.215 MiB, heap total 63.340 MiB, heap used 25.012 MiB, heap profile collection rate 355.329 profiles/s time profile collection rate 0.000 profiles/s
```

aalexand: OK. Is there a comma missing after first "profiles/s"?

ts/src/index.ts
nolanmar511: Fixed

ts/src/index.ts
nolanmar511: fixed.

ts/src/index.ts
nolanmar511: I've added a pause between time and heap now.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -153,22 +153,64 @@ function logError(msg: string, config: Config) {
...
+  // Periodic profiling
+  setInterval(async () => {
     if (!config.disableHeap) {
       const heap = await profiler.profile(
-          {name: 'HEAP-Profile' + new Date(), profileType: 'HEAP'});
+          {name: 'Heap-Profile' + new Date(), profileType: 'HEAP'});
+      heapProfileCount++;
     }
+    await delay(
+        (normalizedConfig.localProfilingPeriodMillis -
```
aalexand: Assuming heap profiles are instantaneous, shouldn't this be just normalizedConfig.localProfilingPeriodMillis / 2? I don't 100% get the subtraction part.

ts/src/config.ts
```diff
@@ -144,5 +162,9 @@ export const defaultConfig = {
   // This is the largest duration for setTimeout which does not cause it to
   // run immediately.
   // https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_settimeout_callback_delay_args.
-  serverBackoffCapMillis: 2147483647
+  serverBackoffCapMillis: 2147483647,
+
+  localProfilingPeriodMillis: 1000,
+  localLogPeriodMillis: 1000,
```
aalexand: IIRC, this was 10000 in the code. Is it intentionally 1000 now?

nolanmar511: Fixed.

ts/src/index.ts
nolanmar511: done.

ts/src/index.ts
aalexand: Looks like it was done.

PR #182, opened 2018-04-30
chore: reduce memory consumption of e2e testing benchmark

nolanmar511:


nolanmar511:
Kokoro failure was a flake in go-get
Re-ran kokoro and it passed.

PR #184, opened 2018-04-30
chore: Update @google-cloud/common to latest version

nolanmar511:


nolanmar511:
Still seeing `Error: Cannot find module '@google-cloud/common'` in circleci and appveyor tests. Can now also reproduce locally (I'd had the module installed in a parent directory before). 

I filed https://github.com/googleapis/nodejs-common/issues/120

nolanmar511:
Switching to `@google-cloud/common` 0.18.X is going to require some work, because the API we use to make requests now uses async/await, and this will change how we deal with getting server-specified backoff.  Hopefully we can still access the part of the response which indicates the backoff.

nolanmar511:
File https://github.com/googleapis/nodejs-common/issues/123

nolanmar511:
Closing. This was done with #187

PR #187, opened 2018-05-01
fix: upgrade to the latest @google-cloud/common

JustinBeckwith:
Once https://github.com/googleapis/nodejs-common/pull/125 lands and we ship `0.18.3`, we should be able to land this one. Apologies again for the trouble!


nolanmar511:
LGTM! (pending 0.18.3 and integration tests passing)

JustinBeckwith:
@nolanmar511 So we cut 18.3.0, and it appears to mostly fix the issues.  Any ideas on what's up with the other failures here?

nolanmar511:
I think it's that `@google-cloud/common` was moved from dependencies to devDependencies. So, the kokoro test doesn't install `@google-cloud/common`.

JustinBeckwith:
🤦‍♂️ good catch.

JustinBeckwith:
@nolanmar511 not sure there's much I can do about the code coverage.  Probably showing up because we deleted a bunch of types 😆

PR #189, opened 2018-05-02
add node 10 to presubmit tests

nolanmar511:


.circleci/config.yml
```diff
@@ -74,6 +76,11 @@ jobs:
       - image: node:9
         user: node
     <<: *unit_tests
+  node10:
+    docker:
+      - image: node:9
```
aalexand: Should be "node:10"?

nolanmar511: Thanks -- fixed.

testing/integration_test.sh
```diff
@@ -26,7 +26,7 @@ cp -R "testing" "$GOPATH/src/proftest"
 cd "$GOPATH/src/proftest"
 go get -t -tags=integration .
 if [ "$KOKORO_GITHUB_PULL_REQUEST_NUMBER" == "" ]; then
-  go test -timeout=30m -parallel=3 -tags=integration -run TestAgentIntegration -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
+  go test -timeout=30m -parallel=4 -tags=integration -run TestAgentIntegration -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
```
aalexand: Not super related to this PR, but: https://golang.org/cmd/go/#hdr-Testing_flags says "The value of this flag is the maximum number of tests to run simultaneously; by default, it is set to the value of GOMAXPROCS." To avoid having to maintain the value here in sync with the number of tests, I would add something like `runtime.GOMAXPROCS(len(testcases))` in the test function instead.

nolanmar511:
PTAL

aalexand:
@nolanmar511 Did you have a successful Kokoro test run?

nolanmar511:
Yes, I did have a successful kokoro run after re-running.

testing/integration_test.sh
aalexand: @nolanmar511 Did you see this comment?

nolanmar511: I switched to using this.

testing/integration_test.go
```diff
@@ -208,7 +209,22 @@ func TestAgentIntegration(t *testing.T) {
...
+				Name:        fmt.Sprintf("profiler-test-node10-%d", runID),
+				MachineType: "n1-standard-1",
+			},
+			name:         fmt.Sprintf("profiler-test-node10-%d-gce", runID),
+			wantProfiles: []profileSummary{{"WALL", "busyLoop"}, {"HEAP", "benchmark"}},
+			nodeVersion:  "10",
+		},
 	}
+
+	// allow test cases to run in parallel.
```
aalexand: Nit: allow -> Allow

Issue #120, opened 2018-01-26
Remove cloud-profiler-nodejs/bindings/statistics.cc as it seems to be unused


aalexand:

nolanmar511:
Waiting to work on #121 first, to see if  #121 might require this.

PR #192, opened 2018-05-05
add node in profile to track external memory

### LINKED ISSUE #157: Buffers don't seem to appear in heap profiles
I ran this code:

```
const profiler = require('@google-cloud/profiler')
profiler.start({
	projectId: 'nolanmar-testproject',
	logLevel: 5,
	serviceContext: {
		service: 'buffer-bench',
		version: '0'
	},
});

var bench;

function buffer_bench() {
    bench = new Buffer(1e6);
    bench.fill('');
    setImmediate(buffer_bench);
}

buffer_bench();
```

And the function buffer_bench didn't appear in the heap profile:
image

The [docs]( https://nodejs.org/api/buffer.html) for Buffer say "Instances of the Buffer class are similar to arrays of integers but correspond to fixed-sized, raw memory allocations outside the V8 heap. The size of the Buffer is established when it is created and cannot be resized."

Does that mean we would expect to not see the full buffer allocation in the profile?
Can this be fixed?

nolanmar511:
Fixes #157 

Testing:

Ran this script:
```
const profiler = require('./out/src/index.js')
profiler.start({
	projectId: 'nolanmar-test',
	logLevel: 5,
	serviceContext: {
		service: 'test-external',
		version: '3'
  },
  // baseApiUrl: 'https://test-cloudprofiler.sandbox.googleapis.com/v2',
});

const startTime = Date.now();
const testArr = [];
const testBuf = [];

/**
 * Fills several arrays, then calls itself with setImmediate.
 * It continues to do this until durationSeconds after the startTime.
 */
function busyLoop(durationSeconds) {
  for (let i = 0; i < testBuf.length; i++) {
    for (let j = 0; j < testBuf[i].length; j++) {
      testBuf[i][j] = Math.sqrt(j * testBuf[i][j]);
    }
  }
  if (Date.now() - startTime < 1000 * durationSeconds) {
    setTimeout(() => busyLoop(durationSeconds), 5);
  }
}

function benchmark(durationSeconds) {
  // Allocate 16 MiB in 64 KiB chunks.
  for (let i = 0; i < 16*16; i++) {
    testBuf[i] = new Buffer(64 * 1024);
  }

  busyLoop(durationSeconds)
}

const durationSeconds = process.argv.length > 2 ? process.argv[2] : 600; 
setTimeout(()=>benchmark(durationSeconds), 1000);
```

This profile, with 16MiB in external, was uploaded:
image

### DIFF
--- a/ts/src/profilers/profile-serializer.ts
+++ b/ts/src/profilers/profile-serializer.ts
@@ -80,7 +80,8 @@ class StringTable {
  */
 function serialize<T extends ProfileNode>(
     profile: perftools.profiles.IProfile, root: T,
-    appendToSamples: AppendEntryToSamples<T>, stringTable: StringTable) {
+    appendToSamples: AppendEntryToSamples<T>, stringTable: StringTable,
+    rootLevelNodes?: Map<string, number[]>) {
   const samples: perftools.profiles.Sample[] = [];
   const locations: perftools.profiles.Location[] = [];
   const functions: perftools.profiles.Function[] = [];
@@ -103,6 +104,18 @@ function serialize<T extends ProfileNode>(
     }
   }
 
+  if (rootLevelNodes) {
+    rootLevelNodes.forEach((value: number[], key: string) => {
+      const f = addFunction(`:${key}`, key, '');
+      const loc = addLocation(`:::${key}`, [new perftools.profiles.Line({
+                                functionId: f.id,
+                              })]);
+
+      samples.push(
+          new perftools.profiles.Sample({locationId: [loc.id], value}));
+    });
+  }
+
   profile.sample = samples;
   profile.location = locations;
   profile.function = functions;
@@ -111,15 +124,19 @@ function serialize<T extends ProfileNode>(
   function getLocation(node: ProfileNode): perftools.profiles.Location {
     const keyStr =
         `${node.scriptId}:${node.lineNumber}:${node.columnNumber}:${node.name}`;
-    let id = locationIdMap.get(keyStr);
+    const id = locationIdMap.get(keyStr);
     if (id !== undefined) {
       // id is index+1, since 0 is not valid id.
       return locations[id - 1];
     }
-    id = locations.length + 1;
+    return addLocation(keyStr, [getLine(node)]);
+  }
+
+  function addLocation(keyStr: string, line: perftools.profiles.Line[]):
+      perftools.profiles.Location {
+    const id = locations.length + 1;
     locationIdMap.set(keyStr, id);
-    const location =
-        new perftools.profiles.Location({id, line: [getLine(node)]});
+    const location = new perftools.profiles.Location({id, line});
     locations.push(location);
     return location;
   }
@@ -133,19 +150,24 @@ function serialize<T extends ProfileNode>(
 
   function getFunction(node: ProfileNode): perftools.profiles.Function {
     const keyStr = `${node.scriptId}:${node.name}`;
-    let id = functionIdMap.get(keyStr);
+    const id = functionIdMap.get(keyStr);
     if (id !== undefined) {
       // id is index+1, since 0 is not valid id.
       return functions[id - 1];
     }
-    id = functions.length + 1;
+    return addFunction(keyStr, node.name || '(anonymous)', node.scriptName);
+  }
+
+  function addFunction(keyStr: string, name: string, scriptName: string):
+      perftools.profiles.Function {
+    const id = functions.length + 1;
     functionIdMap.set(keyStr, id);
-    const nameId = stringTable.getIndexOrAdd(node.name || '(anonymous)');
+    const nameId = stringTable.getIndexOrAdd(name || '(anonymous)');
     const f = new perftools.profiles.Function({
       id,
       name: nameId,
       systemName: nameId,
-      filename: stringTable.getIndexOrAdd(node.scriptName)
+      filename: stringTable.getIndexOrAdd(scriptName)
     });
     functions.push(f);
     return f;
@@ -263,6 +285,7 @@ export function serializeHeapProfile(
         }
       };
 
+
   const stringTable = new StringTable();
   const sampleValueType = createObjectCountValueType(stringTable);
   const allocationValueType = createAllocationValueType(stringTable);
@@ -274,6 +297,10 @@ export function serializeHeapProfile(
     period: intervalBytes,
   };
 
-  serialize(profile, prof, appendHeapEntryToSamples, stringTable);
+  // tslint:disable-next-line: no-any
+  const memoryUsage: {external: number} = process.memoryUsage() as any;
+  serialize(
+      profile, prof, appendHeapEntryToSamples, stringTable,
+      new Map([['(external)', [1, memoryUsage.external]]]));
   return profile;
 }
--- a/ts/test/profiles-for-tests.ts
+++ b/ts/test/profiles-for-tests.ts
@@ -209,15 +209,16 @@ export const v8HeapProfile = {
 
 const heapLines = [
   {functionId: 1, line: 1}, {functionId: 2, line: 5}, {functionId: 3, line: 10},
-  {functionId: 4, line: 8}
+  {functionId: 4, line: 8}, {functionId: 5}
 ];
 
 const heapFunctions = [
   new perftools.profiles.Function({id: 1, name: 5, systemName: 5, filename: 5}),
   new perftools.profiles.Function({id: 2, name: 6, systemName: 6, filename: 7}),
   new perftools.profiles.Function({id: 3, name: 8, systemName: 8, filename: 7}),
+  new perftools.profiles.Function({id: 4, name: 9, systemName: 9, filename: 7}),
   new perftools.profiles.Function(
-      {id: 4, name: 9, systemName: 9, filename: 7})
+      {id: 5, name: 10, systemName: 10, filename: 0})
 ];
 
 const heapLocations = [
@@ -225,6 +226,7 @@ const heapLocations = [
   new perftools.profiles.Location({line: [heapLines[1]], id: 2}),
   new perftools.profiles.Location({line: [heapLines[2]], id: 3}),
   new perftools.profiles.Location({line: [heapLines[3]], id: 4}),
+  new perftools.profiles.Location({line: [heapLines[4]], id: 5}),
 ];
 
 export const heapProfile: perftools.profiles.IProfile = {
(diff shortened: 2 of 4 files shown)


aalexand:
@nolanmar511 The PR description should have a semver prefix?

aalexand:
@nolanmar511 Remove https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/statistics.cc and say that #120 is also fixed? We discussed previously that we'll remove this file if we don't need it to address #157.

ts/src/profilers/profile-serializer.ts
```diff
@@ -111,15 +124,19 @@ function serialize<T extends ProfileNode>(
...
+    const id = locationIdMap.get(keyStr);
     if (id !== undefined) {
       // id is index+1, since 0 is not valid id.
       return locations[id - 1];
     }
-    id = locations.length + 1;
+    return addLocation(keyStr, [getLine(node)]);
+  }
+
+  function addLocation(keyStr: string, line: perftools.profiles.Line[]):
```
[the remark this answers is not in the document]
nolanmar511: Makes sense. 
I ended up creating the fake node and adding it the the v8 profile.

ts/test/profiles-for-tests.ts
```diff
@@ -241,20 +243,14 @@ export const heapProfile: perftools.profiles.IProfile = {
...
-    'objects',
-    'count',
-    'space',
-    'bytes',
-    'main',
-    'function1',
-    'script1',
-    'function3',
-    'function2',
+    '', 'objects', 'count', 'space', 'bytes', 'main', 'function1', 'script1',
```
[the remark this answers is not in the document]
nolanmar511: Yes -- this change came from the linter.

ts/src/profilers/profile-serializer.ts
```diff
@@ -248,6 +248,17 @@ export function serializeTimeProfile(
 export function serializeHeapProfile(
     prof: AllocationProfileNode, startTimeNanos: number,
     intervalBytes: number): perftools.profiles.IProfile {
+  // add node for external memory usage.
+  // tslint:disable-next-line: no-any
```
ofrobots: 1. When disabling a lint rule, add rationale in a comment. E.g. 'Current type definitions do not have external'.
2. Can you submit a PR upstream to fix this? Include a link to the PR here as a TODO here, so that the lint-disable can be removed in the future.

ts/src/profilers/profile-serializer.ts
```diff
@@ -248,6 +248,17 @@ export function serializeTimeProfile(
...
+  const memoryUsage: {external: number} = process.memoryUsage() as any;
```
ofrobots: suggestion: the following might get rid of the unnecessary intermediate `memoryUsage`:

```ts
const {external} : {external: number} = process.memoryUsage() as any;
```

ts/src/profilers/profile-serializer.ts
```diff
@@ -276,4 +287,4 @@ export function serializeHeapProfile(
 
   serialize(profile, prof, appendHeapEntryToSamples, stringTable);
   return profile;
-}
+}
```
ofrobots: nit: add new line.

ts/test/test-helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Makes deep copy of the object o.
+ */
+export function copy<T>(o: T): T {
```
ofrobots: Deep cloning an object in JavaScript is tricky. Just use [a module from npm](https://stackoverflow.com/questions/4459928/how-to-deep-clone-in-javascript), instead of writing your own copy.

ts/test/test-profile-serializer.ts
```diff
@@ -19,36 +19,46 @@ import {serializeHeapProfile, serializeTimeProfile} from '../src/profilers/profi
...
+      heapTotal: 4096,
+      heapUse: 2048,
+    });
   });
   after(() => {
     dateStub.restore();
   });
   it('should produce expected profile', () => {
-    const timeProfileOut = serializeTimeProfile(v8TimeProfile, 1000);
+    const timeProfileOut = serializeTimeProfile(copy(v8TimeProfile), 1000);
```
ofrobots: For my education, why do need to pass a clone to the `serializeTimeProfile` function? Does it mutate its argument?

ts/src/profilers/profile-serializer.ts

ts/src/profilers/profile-serializer.ts

ts/test/test-helpers.ts
nolanmar511: Done.

ts/test/test-profile-serializer.ts
nolanmar511: It's not necessary -- removed this.

ts/src/profilers/profile-serializer.ts
nolanmar511: Made https://github.com/DefinitelyTyped/DefinitelyTyped/pull/25580 to add `external` to `MemoryUsage`.

nolanmar511:
PTAL

ts/test/test-profile-serializer.ts
```diff
@@ -35,20 +44,21 @@ describe('serializeTimeProfile', () => {
...
-    const heapProfileOut = serializeHeapProfile(v8HeapProfile, 0, 512 * 1024);
+    const heapProfileOut =
+        serializeHeapProfile(copy(v8HeapProfile), 0, 512 * 1024);
     assert.deepEqual(heapProfileOut, heapProfile);
   });
   it('should produce expected profile when there is anyonmous function', () => {
-    const heapProfileOut =
-        serializeHeapProfile(v8AnonymousFunctionHeapProfile, 0, 512 * 1024);
+    const heapProfileOut = serializeHeapProfile(
+        copy(v8AnonymousFunctionHeapProfile), 0, 512 * 1024);
```
ofrobots: Is a clone needed here?

ts/test/test-heap-profiler.ts
```diff
@@ -30,12 +31,19 @@ describe('HeapProfiler', () => {
   let stopStub: sinon.SinonStub;
   let profileStub: sinon.SinonStub;
   let dateStub: sinon.SinonStub;
+  let memoryUsageStub: sinon.SinonStub;
   beforeEach(() => {
     startStub = sinon.stub(v8HeapProfiler, 'startSamplingHeapProfiler');
     stopStub = sinon.stub(v8HeapProfiler, 'stopSamplingHeapProfiler');
     profileStub = sinon.stub(v8HeapProfiler, 'getAllocationProfile')
-                      .returns(v8HeapProfile);
+                      .returns(copy(v8HeapProfile));
```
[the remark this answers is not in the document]
nolanmar511: Since, serializeHeapProfile now modifies the heap profile, this is necessary for tests to pass

ts/src/profilers/profile-serializer.ts
```diff
@@ -248,6 +248,19 @@ export function serializeTimeProfile(
 export function serializeHeapProfile(
```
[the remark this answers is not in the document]
aalexand: It's a bit strange the serialization method modifies the profile, now that I think of it. Would it be better to modify the profile as part of the profile() method? https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/33cd53193408353446f3440272ecd858a114f936/ts/src/profilers/heap-profiler.ts#L32

ts/src/profilers/profile-serializer.ts
```diff
@@ -248,6 +248,19 @@ export function serializeTimeProfile(
...
     prof: AllocationProfileNode, startTimeNanos: number,
     intervalBytes: number): perftools.profiles.IProfile {
+  // add node for external memory usage.
```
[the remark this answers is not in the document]

ts/src/profilers/profile-serializer.ts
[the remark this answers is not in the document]
nolanmar511: I agree -- shifted this into heap-profiler.ts.

nolanmar511:
PTAL

ts/test/test-heap-profiler.ts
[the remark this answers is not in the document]
ofrobots: These clones should no longer be necessary, right? If so, can you drop the dependency on `deep-copy` and remove the clones from elsewhere too.

ts/src/profilers/heap-profiler.ts
```diff
@@ -35,6 +36,18 @@ export function profile(): perftools.profiles.IProfile {
...
+  // TODO: remove any once type definition is updated to include external.
+  // tslint:disable-next-line: no-any
+  const {external}: {external: number} = process.memoryUsage() as any;
+  const externalNode: AllocationProfileNode = {
+    name: '(external)',
+    scriptName: '',
+    children: [],
+    allocations: [{sizeBytes: external, count: 1}],
+  };
+  result.children.push(externalNode);
```
ofrobots: I like this a lot better too – thanks.

ts/test/test-heap-profiler.ts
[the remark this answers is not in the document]
nolanmar511: Copying is still needed here, since heapProfiler does modify the returned profile.

nolanmar511:
PTAL

aalexand:
@nolanmar511 I assume the comments were addressed? They are not responded so it's hard to say without looking at the source, but I guess I'll just do that. It's better to explicitly respond to all comments though.

ts/test/test-heap-profiler.ts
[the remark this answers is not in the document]
aalexand: Comment is not added?

ts/src/profilers/profile-serializer.ts
```diff
@@ -80,7 +80,8 @@ class StringTable {
  */
 function serialize<T extends ProfileNode>(
     profile: perftools.profiles.IProfile, root: T,
-    appendToSamples: AppendEntryToSamples<T>, stringTable: StringTable) {
+    appendToSamples: AppendEntryToSamples<T>, stringTable: StringTable,
+    rootLevelNodes?: Map<string, number[]>) {
```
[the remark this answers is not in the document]
nolanmar511: Parameter no longer there.

ts/test/test-profile-serializer.ts
nolanmar511: Done.

ts/test/test-heap-profiler.ts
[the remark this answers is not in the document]
nolanmar511: Done.

nolanmar511:
Sorry -- I'd lost some of the comment chains. 
PTAL.

PR #196, opened 2018-05-10
chore: add language to profile labels

nolanmar511:
Fixes #196 
This change also removes some unused variables from `test-profiler.ts`.


aalexand:
@nolanmar511 the Kokoro build failed?

nolanmar511:
Kokoro build failed on Node10, because the program did not finish profiling before the the tests timed out.
I'm re-running the test, but may also need to look more into why the test timed out on node10.

PR #198, opened 2018-05-15
chore: add language to deployment labels

### LINKED ISSUE #194: Include deployment label language=nodejs in the profiles
The java and go agents do this, we would like to use it for analytics.

nolanmar511:
Fixes #194

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -54,7 +54,7 @@ function isErrorResponseStatusCode(code: number) {
 export interface Deployment {
   projectId?: string;
   target?: string;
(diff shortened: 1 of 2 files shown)


ofrobots:
Can we include the Node.js version and the profiler agent version the labels? It would be good to be able to get this additional data to see use of the module across node versions and across its own versions.

nolanmar511:
Opened https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/issues/200
This would be an easy change, but worth discussing a bit, and tracking before implementing.

PR #201, opened 2018-05-16
chore: stop using deprecated GetCpuProfiler() in node 10

### LINKED ISSUE #106: Use CpuProfiler::New on Node 10 (as v8's GetCpuProfiler() is deprecated starting Node 10)
A new API was introduced in v8/v8@120b753.
The current API is deprecated in V8 6.4: v8/v8@8c5e2d7

We use this method [here](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/time-profiler.cc#L64), [here](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/time-profiler.cc#L70), [here](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/time-profiler.cc#L78), and [here](https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/master/bindings/time-profiler.cc#L83)

Node.js may also have changes in response to this. https://github.com/nodejs/node/issues/18039

nolanmar511:
Fixes #106

### DIFF
--- a/bindings/time-profiler.cc
+++ b/bindings/time-profiler.cc
@@ -19,6 +19,15 @@
 
 using namespace v8;
 
+#if NODE_MODULE_VERSION > NODE_8_0_MODULE_VERSION
+// This profiler exists for the lifetime of the program. Not calling 
+// CpuProfiler::Dispose() is intentional.
+CpuProfiler* cpuProfiler = CpuProfiler::New(v8::Isolate::GetCurrent());
+#else
+CpuProfiler* cpuProfiler = v8::Isolate::GetCurrent()->GetCpuProfiler();
+#endif
+
+
 Local<Value> TranslateTimeProfileNode(const CpuProfileNode* node) {
   Local<Object> js_node = Nan::New<Object>();
   js_node->Set(Nan::New<String>("name").ToLocalChecked(),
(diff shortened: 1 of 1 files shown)


bindings/time-profiler.cc
```diff
@@ -19,6 +19,13 @@
 
 using namespace v8;
 
+#if (NODE_MODULE_VERSION > NODE_8_0_MODULE_VERSION)
```
aalexand: Parens are not needed.

bindings/time-profiler.cc
```diff
@@ -19,6 +19,13 @@
...
+CpuProfiler* cpuProfiler = CpuProfiler::New(v8::Isolate::GetCurrent());
```
aalexand: I feel you maybe got a memory leak. Don't you need to dispose the profiler with the new flow? The name change from Get() -> New() implies some changes in how the life time is managed I think. I also think you want to retain it between the profiling sessions to avoid re-creating it.

ofrobots: The variable is at file global scope, so the profiler is being retained across the profiling sessions. The lack of a [`CpuProfiler::Dispose`](https://cs.chromium.org/chromium/src/v8/include/v8-profiler.h?type=cs&sq=package:chromium&g=0&l=303) may seem inelegant, but I wouldn't call it a leak. The profiler exists for the lifetime of the application.

ofrobots: The question I have is what happens if there are multiple modules in the system that create the CPU profiler using `CpuProfiler::New`. Would both of them end up working correctly? /cc @a1ph

aalexand: Apologies, I missed that it's in global scope. I'd add a comment that the leak is intended.

aalexand: By the way, completely aside but: we ignore the possibility of existence of multiple isolates, right?

a1ph: You can create as many profilers as you need. You can even have several for the same isolate. We have a test for that https://cs.chromium.org/chromium/src/v8/test/cctest/test-cpu-profiler.cc?type=cs&q=MultipleProfilers&sq=package:chromium&g=0&l=2472

bindings/time-profiler.cc
nolanmar511: Done.

bindings/time-profiler.cc
nolanmar511: Added a comment stating that `Dispose` is not called intentionally.

nolanmar511:
PTAL

nolanmar511:
PTAL

PR #203, opened 2018-05-17
feat: get service from gke cluster name

### LINKED ISSUE #81: Auto-discover service on GKE
@google-cloud/cloud-debug-nodejs uses the cluster name for the service (https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/a828ea62a434e3c92eda46e39921531ba30e68cf/src/agent/debuglet.ts#L347). Profiler should do the same.

nolanmar511:
Fixes #81

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -45,12 +45,33 @@ function hasService(config: Config):
       typeof config.serviceContext.service === 'string';
 }
 
+interface LocalProfilerConfig extends Config {
+  projectId?: string;
+  logLevel: number;
+  serviceContext: {service?: string; version?: string;};
+  instance?: string;
+  zone?: string;
+  disableTime: boolean;
+  disableHeap: boolean;
+  timeIntervalMicros: number;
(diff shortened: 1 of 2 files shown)


ts/src/index.ts
```diff
@@ -118,6 +136,17 @@ async function initConfigMetadata(config: ProfilerConfig):
       config.instance = instance;
     }
   }
+  if (!config.serviceContext.service) {
+    try {
+      config.serviceContext.service =
+          await getMetadataInstanceField('attributes/cluster-name');
```
aalexand: I am not convinced this is useful. The cluster typically runs many microservices (pods in Kubernetes speak). Aggregating all of them together is not useful.

ofrobots: Is there a way to figure out the pod name? I agree that would be better, and we should probably fix that in the debug agent as well.

nolanmar511:
Closing this  -- I don't know that we can easily get the pod names, and using cluster names could be confusing.

PR #208, opened 2018-05-22
chore: retry flaky steps in e2e test

nolanmar511:


testing/integration_test.go
```diff
@@ -50,35 +50,42 @@ const startupTemplate = `
...
 # Fail on any error
 set -eo pipefail
 
 # Display commands being run
 set -x
 # Install git
-apt-get update >/dev/null
-apt-get -y -q install git build-essential  >/dev/null
+retry apt-get update >/dev/null
+retry apt-get -y -q install git build-essential  >/dev/null
```
aalexand: Nit: double-space before ">/dev/null".

testing/integration_test.go
```diff
@@ -50,35 +50,42 @@ const startupTemplate = `
...
 cd cloud-profiler-nodejs
-git fetch origin {{if .PR}}pull/{{.PR}}/head{{else}}{{.Branch}}{{end}}:pull_branch
+retry git fetch origin {{if .PR}}pull/{{.PR}}/head{{else}}{{.Branch}}{{end}}:pull_branch
 git checkout pull_branch
 git reset --hard {{.Commit}}
-npm install >/dev/null
-npm run compile
-npm pack >/dev/null
+retry npm install >/dev/null
+retry npm run compile
```
aalexand: I feel this command is purely local?

testing/integration_test.go
```diff
@@ -50,35 +50,42 @@ const startupTemplate = `
...
+retry npm pack >/dev/null
```
aalexand: I feel this command is purely local?

testing/integration_test.go
```diff
@@ -87,7 +94,7 @@ mkdir -p "$TESTDIR"
 cp "testing/busybench.js" "$TESTDIR"
 cd "$TESTDIR"
 
-npm install "$PROFILER" >/dev/null
+retry npm install "$PROFILER" >/dev/null
```
aalexand: This installs local archive so shouldn't have any network interactions? Or it has for deps?

testing/integration_test.go
nolanmar511: Yes, this is -- fixed.

testing/integration_test.go
nolanmar511: Fixed.

testing/integration_test.go
nolanmar511: I think this is local. Changing this.

testing/integration_test.go
nolanmar511: I think this is local, but we've also had a flake here. 
I think it makes sense to retry this and npm pack together.

nolanmar511: We've had a flake with `npm install "$PROFILER"` (continuous build 899). But when looking at that flake, I don't know that can tell whether is occurred because npm pack did something bad and so we tried to install a malformed file, or if it occurred because something went wrong with npm install.

nolanmar511: Looked more into the flake from continuous build 899 -- I think that the problem was in the `npm install` step, but that retrying likely would not have helped. Generally calling npm install will initialize a package.json if there isn't one. I'm guessing there may have been a problem with npm install which changed this behavior during this test. 

Done.

testing/integration_test.go
```diff
@@ -50,44 +50,55 @@ const startupTemplate = `
...
-PROFILER="$HOME/cloud-profiler-nodejs/google-cloud-profiler-$VERSION.tgz"
 
 TESTDIR="$HOME/test"
 mkdir -p "$TESTDIR"
 cp "testing/busybench.js" "$TESTDIR"
 cd "$TESTDIR"
 
-npm install "$PROFILER" >/dev/null
+packAndInstall() {
+	npm pack $HOME/cloud-profiler-nodejs >/dev/null
```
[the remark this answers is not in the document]
nolanmar511: Removed this retry after further investigating the flake which had caused me to add this.

PR #223, opened 2018-06-06
fix: coerce node version so pre, nightly, alpha, and beta versions of node can be used with profiler.

### LINKED ISSUE #221: Not possible to run this with Node.js prerelease/nightly binaries
While investigating #199, I ran into this issue:

```sh
$ ~/src/node/node --require . testing/busybench.js 600
ERROR:@google-cloud/profiler: Error: Could not start profiler: node version v11.0.0-pre does not satisfies "6.12.3 - 7.0.0 || >=8.9.4"
```

That semver test prevents the profiler from running against Node.js nightly or pre-release build. IMO, this should be a warning rather than a fatal error.

nolanmar511:
Fixes #221.

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -127,11 +127,13 @@ async function initConfigMetadata(config: ProfilerConfig):
  * rejected promise.
  */
 export async function createProfiler(config: Config): Promise<Profiler> {
-  if (!semver.satisfies(process.version, pjson.engines.node)) {
-    logWarning(
-        `Node version ${process.version}` +
-            ` does not satisfies "${pjson.engines.node}"`,
-        config);
+  // Coerce version if possible, to remove any pre-release, alpha, beta, etc
+  // tags.
+  const version = semver.coerce(process.version) || process.version;
+  if (!semver.satisfies(version, pjson.engines.node)) {
+    throw new Error(
+        `Could not start profiler: node version ${process.version}` +
+        ` does not satisfies "${pjson.engines.node}"`);
   }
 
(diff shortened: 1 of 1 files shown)


ts/src/index.ts
```diff
@@ -128,9 +128,10 @@ async function initConfigMetadata(config: ProfilerConfig):
  */
 export async function createProfiler(config: Config): Promise<Profiler> {
   if (!semver.satisfies(process.version, pjson.engines.node)) {
-    throw new Error(
-        `Could not start profiler: node version ${process.version}` +
-        ` does not satisfies "${pjson.engines.node}"`);
+    logWarning(
```
aalexand: Can we make it a warning (or just skip the check) for versions with pre-release tag, but keep it an error for released versions?

ofrobots:
I would prefer if this was kept as a warning. A real error will be thrown
if the version is not compatible. Let's not over complicate things.

aalexand:
> A real error will be thrown if the version is not compatible

@ofrobots This check is to mitigate users hitting known runtime issues (sporadic crash in heap profiler for Node 6, memory leak in CPU profiler for Node 8, IIRC), no error will be thrown, but user's production app will misbehave. We should add a similar check for Node 10 to disallow using heap profiler until the bug with weak references not garbage collected as expected. I think these checks should be an error so that users do not hit known issues wasting time figuring out issues in their production apps.

@nolanmar511 Did you look into why Circle CI is failing?

nolanmar511:
PTAL -- fixed CI (I had a formatting error)

aalexand:
LGTM

PR #224, opened 2018-06-08
chore: add retries to e2e test

nolanmar511:


nolanmar511:
This is also added to go agent tests and java agent tests (jianqiao approved those).

Issue #188, opened 2018-05-02
Have an e2e test against a node version that uses v8 very close to head


aalexand:
As CPU or heap profiler in v8 get any changes, we should continuously make sure those do not carry surprises for us (e.g. there is https://chromium-review.googlesource.com/c/v8/v8/+/1039825, which shouldn't break us but only green test gives confidence). To do that we need an automated e2e test run against a version of node that uses v8 very close to head. @ofrobots any recommendations on how to organize this? My naive hope would be that we could do something like "nvm install head" and the head gets v8 updated regularly (automatically?).

@nolanmar511 FYI

ofrobots:
This is a good idea. Having feedback from potentially breaking changes in V8 as early as possible would be great. I am not sure whether the tests we have in the Node project are as extensive for the profiling/diagnostics use-cases, and this project might well be one of the most advanced users of these V8 APIs.

In the node project we run `v8-canary` builds which are supposed to be Node.js master integrated with a recent (within O(days) of V8 Tip-of-Tree) version of V8. Available [here](https://nodejs.org/download/v8-canary/). We can download and unpack the tarball and go from there. I think this would be sufficient for our purposes.

If we want a build that is within O(hours) up to date, we can use the tarballs from the V8 integration build: https://uberchromegw.corp.google.com/i/client.v8.fyi/builders/V8%20Linux64%20-%20node.js%20integration. This build integrates each V8-roll (small number of commits) into a semi-recent Node.js snapshot. The built binaries are available for download as a tarball from that site too. This will provide semi-immediate feedback if a V8 commit ends up breaking us.

aalexand:
O(days) is good but O(hours) is better so I'd go for that.

nolanmar511:
Installing the v8-canary builds can be done in one line:
```
NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/v8-canary/ nvm i node
```

Installing the v8 integration build would be harder. There's some javascript code which does this, but it first retrieves retrieves a list of builds, then finds the latest successful build, then gets the build git revision and commit and then downloads the build. I don't think we'll need to get the revision and commit (those seem to be used to in reporting). 
I think we'd need maybe ~100 lines of additional code to use the integration build, which isn't huge. Would we want to duplicate the javascript code and call it from the start-up script? Translate it to go?

aalexand:
[...]

PR #226, opened 2018-06-11
chore: run e2e test with v8 canary build

nolanmar511:
Starts running e2e tests against v8 canary build.
For #188; still looking into running against v8 integration build.


nolanmar511:
Closing -- since we need to get header files, it makes more sense to use v8-lkgr build

PR #248, opened 2018-07-06
fix: distinguish idle from program time

ofrobots:
Fixes: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/issues/38


ts/src/profilers/time-profiler.ts
```diff
@@ -35,10 +35,20 @@ export class TimeProfiler {
    * @param durationMillis - time in milliseconds for which to collect profile.
    */
   async profile(durationMillis: number): Promise<perftools.profiles.IProfile> {
+    // Node.js contains an undocumented API for reporting idle status to V8.
+    // This lets the profiler distinguish idle time from time spent in native
+    // code. Ideally this should be default behavior. Until then, use the
+    // undocumented API.
+    // See https://github.com/nodejs/node/issues/19009#issuecomment-403161559.
+    // tslint:disable-next-line no-any
+    (process as any)._startProfilerIdleNotifier();
```
aalexand: Does this API exist on all Node.js versions? Should this call and the call below check whether the API exists to avoid failing the profiling if it does not?

ofrobots: It exists in all versions supported by this module.

aalexand:
@ofrobots The CI fails on node 6?

ofrobots:
I've relaunched the job. The error message didn't seem related. /cc @nolanmar511 in case you have an idea of what is going on.

aalexand:
@ofrobots Can this be merged?

ofrobots:
There is no LGTM, but this is good to go from my side :).

aalexand:
@ofrobots Ah, I thought I LGTM'ed it. Done now. Please merge once CI has finished.

PR #249, opened 2018-07-11
fix: parse the response body message to determine retry backoff

### LINKED ISSUE #245: Server-specified backoff is broken
Starting with `@google-cloud/common` 0.18.X, the response returned when there is an error is null, so we don't have the portion of the response needed to identify the server-specified backoff. 

I've filed https://github.com/googleapis/nodejs-common/issues/172

nolanmar511:
This fixes #245.

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -81,16 +81,16 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
     undefined {
   // tslint:disable-next-line: no-any
   const body = (response as any).body;
-  if (body && body.error && body.error.details &&
-      Array.isArray(body.error.details)) {
-    for (const item of body.error.details) {
-      if (typeof item === 'object' && item.retryDelay &&
-          typeof item.retryDelay === 'string') {
-        const backoffMillis = parseDuration(item.retryDelay);
-        if (backoffMillis > 0) {
-          return backoffMillis;
-        }
-      }
+
+  // The response currently does not have field containing the retry duration.
+  // As a work-around, response body's message is parsed to get the backoff 
+  // duration.
+  // TODO: Remove this work-around and get the retry delay from
+  // body.error.details.
+  if (body && body.message && typeof body.message === 'string') {
+    const backoffMillis = parseDuration(body.message);
+    if (backoffMillis > 0) {
+      return backoffMillis;
     }
   }
   return undefined;
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -73,9 +73,6 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
-const BACKOFF_MSG_PAT =
-    /action throttled, backoff for ((?:([0-9]+)h)?(?:([0-9]+)m)?([0-9.]+)s)$/;
-
 /**
  * @return number indicated by backoff if the response indicates a backoff and
  * that backoff is greater than 0. Otherwise returns undefined.
@@ -102,10 +99,13 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
  *
  * Public for testing.
  */
-export function parseBackoffDuration(backoffStr: string): number|undefined {
-  const found = backoffStr.match(BACKOFF_MSG_PAT);
-  if (found && found.length >= 2 && typeof found[1] === 'string') {
-    const backoffMillis = parseDuration(found[1]);
+export function parseBackoffDuration(backoffMessage: string): number|undefined {
+  const backoffMessageRegex =
+      /action throttled, backoff for ((?:([0-9]+)h)?(?:([0-9]+)m)?([0-9.]+)s)$/;
+  const [, duration] =
+      backoffMessageRegex.exec(backoffMessage) || [undefined, undefined];
+  if (duration) {
+    const backoffMillis = parseDuration(duration);
     if (backoffMillis > 0) {
       return backoffMillis;
     }
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -73,6 +73,9 @@ export interface RequestProfile {
   labels?: {instance?: string};
 }
 
+const BACKOFF_MSG_PAT =
+    /action throttled, backoff for ((?:([0-9]+)h)?(?:([0-9]+)m)?([0-9.]+)s)$/;
+
 /**
  * @return number indicated by backoff if the response indicates a backoff and
  * that backoff is greater than 0. Otherwise returns undefined.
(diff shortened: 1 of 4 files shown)


ts/test/test-profiler.ts
```diff
@@ -637,7 +637,9 @@ describe('Profiler', () => {
                            .onCall(0)
                            .callsArgWith(1, undefined, undefined, {
                              statusCode: 409,
-                             body: {error: {details: [{retryDelay: '50s'}]}}
+                             body: {
+                               message: 'action throttled, backoff for 50s',
```
aalexand: Add a test for a string like "1m2s"? The backoff duration in the message is in the form "72h3m0.5s".

aalexand: Here are test cases I used elsewhere:

```java
    assertThat(tryParseBackoffDuration("")).isEqualTo(0.0);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 1m2s")).isEqualTo(62.0);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 2.5s")).isEqualTo(2.5);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 1m2.5s")).isEqualTo(62.5);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 10m0s")).isEqualTo(600.0);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 1h2.5s")).isEqualTo(3602.5);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 1h1m2.5s")).isEqualTo(3662.5);
    assertThat(tryParseBackoffDuration("action throttled, backoff for 1m2+s")).isEqualTo(0.0);
```

ts/src/profiler.ts
```diff
@@ -81,16 +81,16 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
...
-        const backoffMillis = parseDuration(item.retryDelay);
-        if (backoffMillis > 0) {
-          return backoffMillis;
-        }
-      }
+
+  // The response currently does not have field containing the retry duration.
+  // As a work-around, response body's message is parsed to get the backoff 
+  // duration.
+  // TODO: Remove this work-around and get the retry delay from
```
aalexand: Please use consistent terminology. Currently you use "retry duration" vs. "backoff duration" vs. "retry delay" in this documentation paragraph.

Also: work-around -> workaround.

ts/test/test-profiler.ts
nolanmar511: Added test cases matching these.

ts/src/profiler.ts

nolanmar511:
PTAL

ts/src/profiler.ts
```diff
@@ -81,16 +84,30 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
...
-        const backoffMillis = parseDuration(item.retryDelay);
-        if (backoffMillis > 0) {
-          return backoffMillis;
-        }
-      }
+
+  // The response currently does not have field containing the server-specified
+  // backoff. As a workaround, response body's message is parsed to get the
+  // backoff.
+  // TODO: Remove this work-around and get the retry delay from
```
aalexand: Nit: you still have the "work-around" rather than "workaround" spelling here. Could you change it here as well (and elsewhere, if any)?

aalexand: Perhaps link to the Github issue that addressing this TODO depends on?

aalexand: Sorry, I don't see a link added in the code - am I missing it?

ts/src/profiler.ts
```diff
@@ -81,16 +84,30 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
...
+
+/**
+ * @return if the backoff duration can be parsed, then the backoff duration in
+ * ms, otherwise undefined.
+ *
+ * Public for testing.
+ */
+export function parseBackoffDuration(backoffStr: string): number|undefined {
+  const found = backoffStr.match(BACKOFF_MSG_PAT);
+  if (found && found.length >= 2 && typeof found[1] === 'string') {
```
ofrobots: You could write this as follows. IMO this is more readable.

```ts
const [, duration] = BACKOFF_MSG_PATH.exec(backoffStr);
```

Other nits:
* The parameter `parameterStr` include the (abbreviated) type in the name. It would be better to call it `backoffMessage` or just `backoff`.
* Personal opinion: the regex being declared at the top of the file makes it harder to understand what this code is doing. I would suggest putting the regex (or the variable) in this function.

aalexand: Putting the regexp var within the function would make it instantiate on every call?

third_party/gtest/googlemock/configure.ac
```diff
@@ -139,7 +139,7 @@ AS_IF([test "x${HAVE_BUILT_GTEST}" = "xyes"],
       GTEST_LIBS='$(top_builddir)/../googletest/lib/libgtest.la'
       GTEST_VERSION="${GTEST_MIN_VERSION}"])
 
-# TODO(chandlerc@google.com) Check the types, structures, and other compiler
+# (chandlerc@google.com) Check the types, structures, and other compiler
```
aalexand: This change seems unrelated.

ts/src/profiler.ts
```diff
@@ -81,16 +81,34 @@ function getServerResponseBackoff(response: http.IncomingMessage): number|
...
+  // TODO (issue #250): Remove this workaround and get the retry delay from
+  // body.error.details.
+  if (body && body.message && typeof body.message === 'string') {
+    return parseBackoffDuration(body.message);
+  }
+  return undefined;
+}
+
+
+const BACKOFF_MSG_PATH =
```
aalexand: Why is it "PATH"? Did you mean "_PAT" for "pattern"? Maybe use "_REGEX", I think it would be a bit cleaner.

ts/src/profiler.ts
nolanmar511: For now, I moved the regex closer to the function, but I'd be happy to put the regex in the function if it won't instantiate on every call.

ts/src/profiler.ts

third_party/gtest/googlemock/configure.ac
nolanmar511: Done.

nolanmar511:
PTAL

ts/src/profiler.ts
ofrobots: I would call this premature optimization :p.

There is marginal difference in performance: http://jsben.ch/jsLlz.

ofrobots: Actually there is no statistically significant difference in performance.

aalexand:
@nolanmar511 The CI test failed on node10?

aalexand:
@nolanmar511 Please merge once node10 CI is passing. (assuming ofrobots@ doesn't have more comments)

PR #255, opened 2018-07-18
test: use strictEqual in tests

JustinBeckwith:


nolanmar511:
Tests are failing because of formatting.
Could you run `npm run fix`?

Issue #19, opened 2017-10-20
Time profile sample / call site locations should carry accurate line number


nolanmar511:
Check that profile line numbers are provided by v8 profiler and make sure that (if provided) line number is in serialized profile

aalexand:
@ofrobots We would like to have accurate per-source-line locations in the profiles, is this possible?

aalexand:
@nolanmar Please also take a look at whether we include parent lexical function name into the (anonymous) function names in the profiles

aalexand:
@ofrobots I think you mentioned there is an ongoing discussion on this, can you link that here?

ofrobots:
@fhinkel is there a tracking bug for making the profiler more accurate work you're planning to do?

fhinkel:
I think this one https://bugs.chromium.org/p/v8/issues/detail?id=6239 is relevant

ofrobots:
Thanks. Hmm.. that one isn't precise enough on the precise requirements. @aalexand, @nolanmar511 I will setup a meeting with the relevant folks.

fhinkel:
The V8 CpuProfiler provides more exact line numbers. But the cloud-profiler doesn't use them.  

Every node  has `LineTicks`. There are `GetHitLineCount` many `LineTick` entries. Every tick has a `line` and a `hit_count`. These line numbers are as precise as we can be for code objects.

```cpp

Local<Value> TranslateTimeProfileNode(const CpuProfileNode* node) {
  //...
  unsigned int hitLineCount = node->GetHitLineCount();
  std::vector<CpuProfileNode::LineTick> entries(hitLineCount);
  if (node->GetLineTicks(&entries[0], hitLineCount)) {
    for (const CpuProfileNode::LineTick entry: entries) {
      printf("  line %d, ticks %d in %s\n", entry.line, entry.hit_count, node->GetScriptResourceNameStr());
    }
  }
// ...
```


The `GetLineNumber` of a  `CpuProfileNode` matches the beginning of the function. The inlining information is used to provide the correct function, not what it was inlined to. The ticks show the line numbers, where the sample was taken, so for inline functions they look off. 

I wrote up a proof-of-concept that takes the ticks into consideration during the translation step: 
https://github.com/fhinkel/cloud-profiler-nodejs/commit/1789008b879bd1709d050eff4539032a1438c76d
In the PoC, `pprof` shows nodes for several line numbers in the same function. 

```
[...]

PR #258, opened 2018-07-25
feat: make it possible to collect line-accurate profiles.

nolanmar511:
Part of #19

DO NOT SUBMIT.
This change is ready to be reviewed, but we should not submit until #348 and #349 are fixed (or, at least, much more understood and documented).


nolanmar511:
PTAL

ts/src/profilers/profile-serializer.ts
```diff
@@ -209,7 +235,8 @@ function createAllocationValueType(table: StringTable):
  * @param intervalMicros - average time (microseconds) between samples.
  */
 export function serializeTimeProfile(
-    prof: TimeProfile, intervalMicros: number): perftools.profiles.IProfile {
+    prof: TimeProfile, intervalMicros: number,
```
ofrobots: nit: add `hasLines` to the jsdoc comment above as well.

ts/src/profilers/profile-serializer.ts
```diff
@@ -81,7 +82,7 @@ class StringTable {
 function serialize<T extends ProfileNode>(
     profile: perftools.profiles.IProfile, root: T,
     appendToSamples: AppendEntryToSamples<T>, stringTable: StringTable,
-    ignoreSamplesPath?: string) {
+    hasLines: boolean, ignoreSamplesPath?: string) {
```
ofrobots: nit: add `hasLines` to the jsdoc comment above as well.

ts/src/profilers/time-profiler.ts
```diff
@@ -23,7 +23,8 @@ export class TimeProfiler {
   /**
    * @param intervalMicros - average time in microseconds between samples
    */
-  constructor(private intervalMicros: number) {
+  constructor(
+      private intervalMicros: number, private profileLineNumbers: boolean) {
```
ofrobots: nit: adjust jsdoc comment above.

ts/src/profilers/profile-serializer.ts

ts/src/profilers/profile-serializer.ts

ts/src/profilers/time-profiler.ts
nolanmar511: Done.

nolanmar511:
PTAL

bindings/profiler.cc
```diff
@@ -85,73 +82,142 @@ NAN_METHOD(StopSamplingHeapProfiler) {
...
+                  Nan::New<Number>(profile->GetStartTime()));
   js_profile->Set(Nan::New<String>("endTime").ToLocalChecked(),
-    Nan::New<Number>(profile->GetEndTime()));
+                  Nan::New<Number>(profile->GetEndTime()));
   return js_profile;
 }
 
 NAN_METHOD(StartProfiling) {
-  Local<String> name = info[0].As<String>();
+  if (info.Length() != 2) {
```
ofrobots: I find it useful to have a comment as the first line indicating the JS function signature so that reading the code is easier. Based on the code below, I expect the comment to be
```c++
// startProfiling(name: String, includeLineInfo: boolean)
```

bindings/profiler.cc
```diff
@@ -85,73 +82,142 @@ NAN_METHOD(StopSamplingHeapProfiler) {
...
+  }
+  if (!info[0]->IsString()) {
+    return Nan::ThrowTypeError("First argument type must be a string.");
+  }
+  if (!info[1]->IsBoolean()) {
+    return Nan::ThrowTypeError("First argument type must be a boolean.");
+  }
+  Local<String> name =
+      Nan::MaybeLocal<String>(info[0].As<String>()).ToLocalChecked();
+  bool includedLineInfo =
```
ofrobots: Is the expectation that the caller must pass the same value for `includedLineInfo` as what they passed to `StartProfiling`. What happens when they pass in the wrong value here?

It might be better to keep this state on the C++ side and not require the caller to pass the same value to `Start` and `StopProfiling`.

bindings/profiler.cc
```diff
@@ -85,73 +82,142 @@ NAN_METHOD(StopSamplingHeapProfiler) {
...
   int32_t count = node->GetChildrenCount();
-  Local<Array> children = Nan::New<Array>(count);
+
+  unsigned int index = 0;
+  Local<Array> children;
+
+  // Add nodes corresponding to lines within the node's function.
+  unsigned int hitLineCount = node->GetHitLineCount();
+  std::vector<CpuProfileNode::LineTick> entries(hitLineCount);
+  if (hasLines && node->GetLineTicks(&entries[0], hitLineCount)) {
```
ofrobots: Can `node->GetLineTicks(..)` return true when `hasLines` is `false`? If not, do we even need that to be a configurable parameter to this function?

bindings/profiler.cc
```diff
@@ -85,73 +82,142 @@ NAN_METHOD(StopSamplingHeapProfiler) {
...
+    children = Nan::New<Array>(count + entries.size());
+    js_node->Set(Nan::New<String>("hitCount").ToLocalChecked(),
```
ofrobots: A comment here would be good to explain what is going on and why the number 0 is special.

bindings/profiler.cc
nolanmar511: Done.

bindings/profiler.cc
nolanmar511: There is the expectation that the same values are passed in. 

If, for example, `includeLineInfo=true` for `StartProfiling`, and `false` for `StopProfiling`, then LineTick information would not be included in the returned profile, so it wouldn't be possible to get accurate line numbers for leaf nodes. 

I guess this variable could be renamed to "includeLineTicks" in stop profiling. 

I'm not sure about storing this as state, since we'd have to basically store a map of name used to start profile to that boolean value, and I'd be worried about not deleting things from the array properly.

bindings/profiler.cc
nolanmar511: This will always be true.

When there are detailed line numbers, the line number of a node refers to the line from which it was called, so it's necessary to get leaf node's line information from the LineTicks array.

bindings/profiler.cc
nolanmar511: Sorry -- getting confused with github.

Which line does this comment refer to?

ofrobots: `hasDetailedLines && node->GetLineTicks(&entries[0], hitLineCount`

bindings/profiler.cc
ofrobots: Do we have test cases that exercise incorrect usage of the API (and ensure that the failure case isn't anything bad). That is, a test case that passes in mismatched value for `includeLineInfo` between start and stop.

sduskis:
@nolanmar511, this PR has gone stale. Is it worth reviving?

aalexand:
We want to enable tests for line numbers Soo enough though to ensure no
regressions in the underlying implementation.

nolanmar511:
Closing -- this change should now be made to https://github.com/google/pprof-nodejs.

PR #262, opened 2018-07-30
feat: hide samples with path containing @google-cloud/profiler by default

### LINKED ISSUE #205: Hiding heap profiler internals in profiles
Should we hide profiler internals from the profiles?
There's some evidence that showing these is confusing, and (unlike other code) profiling-related code runs when profiles are collected (so cannot be randomly sampled by the profiler).

nolanmar511:
This shows the diff between profiles collected before and after this change. These profiles are on the project nolanmar-test-1 (which I think I've added aalexand and ofrobots to as viewers -- let me know if this isn't the case).

image

Fixes #205

### DIFF
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -78,8 +78,8 @@ export interface Config extends GoogleAuthOptions {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
-  // Samples with stacks containing this substring will not be included in
-  // heap profiles.
+  // Samples with stacks with any location containing this as a substring
+  // in their file name will not be included in heap profiles.
   // By default this is set to "@google-cloud/profiler" to exclude samples from
   // the profiler.
   ignoreHeapSamplesPath?: string;
--- a/ts/src/config.ts
+++ b/ts/src/config.ts
@@ -78,6 +78,12 @@ export interface Config extends GoogleAuthOptions {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
+  // Samples with stacks containing this substring will not be included in
+  // heap profiles.
+  // By default this is set to "@google-cloud/profiler" to exclude samples from
+  // the profiler.
+  ignoreHeapSamplesPath?: string;
+
   // On each consecutive error in profile creation, the backoff envelope will
   // increase by this factor. The backoff will be a random value selected
   // from a uniform distribution between 0 and the backoff envelope.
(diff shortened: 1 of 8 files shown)


ts/src/config.ts
```diff
@@ -78,6 +78,12 @@ export interface Config extends GoogleAuthOptions {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
+  // Samples with stacks containing this substring will not be included in
+  // heap profiles.
+  // By default this is set to "@google-cloud/profiler" to exclude samples from
+  // the profiler.
+  ignoreHeapSamplesPath?: string;
```
aalexand: Do we really need this to be configurable? Is this to be able to turn this off during debugging?

ts/src/config.ts
```diff
@@ -78,6 +78,12 @@ export interface Config extends GoogleAuthOptions {
   // stack depth may increase overhead of profiling.
   heapMaxStackDepth?: number;
 
+  // Samples with stacks containing this substring will not be included in
```
aalexand: "stacks containing this substring" is unclear. What exact attribute is matched?

ts/src/profilers/heap-profiler.ts
```diff
@@ -41,14 +42,17 @@ export function profile(): perftools.profiles.IProfile {
...
   const {external}: {external: number} = process.memoryUsage() as any;
-  const externalNode: AllocationProfileNode = {
-    name: '(external)',
-    scriptName: '',
-    children: [],
-    allocations: [{sizeBytes: external, count: 1}],
-  };
-  result.children.push(externalNode);
-  return serializeHeapProfile(result, startTimeNanos, heapIntervalBytes);
+  if (external > 0) {
```
aalexand: This seems like an independent change?

ts/src/config.ts
nolanmar511: Yes, this is to be able to turn this off for debugging. 
Should I switch it to a boolean? It might make initializing this string trickier, but it can be done.

ts/src/profilers/heap-profiler.ts
nolanmar511: Sorry. I made this change to simplify testing, so I wouldn't have to add an "(external)" function and location to the profiles I added. 
Should I undo this?

ts/src/config.ts
nolanmar511: Tried to clarify this.

aalexand:
@nolanmar511 The appveyor and npmaudit CI steps are failing - PTAL? You can merge once those are green.

PR #268, opened 2018-08-07
feat: add source map support

### LINKED ISSUE #204: Add source map support
Right now, the profiler reports javascript functions and line numbers. This makes it harder to interpret profiles for users who have written code in typescript. 

We should add source map support so that profiles can report typescript line numbers and function names.

nolanmar511:
Part of #204.

With this change, the profiling agent will recursively search all directories in the working directory (or other specified directories), excluding any directories named node_modules or .git, and find the source maps within the working directory (or other specified directories).

These source maps will then be used to map generated code locations in heap and cpu profiles to source code locations in the original code base.

Performance measurements (on an N1 small GCE VM):
Running busybench used in e2e tests (This kind of represents a lower bound :
* No source maps
   * Over a 10 minute run:
     * Avg. heap serialization:  0.000323s (std dev 0.0000770)
     * Avg cpu serialization: 0.000489s (std dev 0.000349)
* No dependency source maps (this change -- fairly similar to when there are no source maps)
   *  1 map file (for busybench.ts)
   *  0.0097s to create source mapper.
   * Over a 10 minute run:
     * Avg. heap serialization:  0.000116s (std dev 0.00138)
     * Avg cpu serialization: 0.000689s (std dev 0.000928)
* Searching all dirs (including node_modules) for source maps (serializing profiles takes ~10x as long compared to no source maps)
   *  1147 map files
   *  0.757820275s to create source mapper.
   * Over a 10 minute run:
     * Avg. heap serialization:  0.00182s (std dev 0.00197)
     * Avg cpu serialization: 0.00329s (std dev 0.00729)

I wasn't sure what a good representative typescript app would be -- particularly since I think the number of source maps is important. But I can re-run with an express app that sends traffic to itself if more testing is needed (and plan to re-run with an express app as I implement lazy loading).

Plan going forward:
We would like to make it possible for users to automatically use source maps from node_modules. To allow for this without increasing the time to start profiling, we want to lazily read in source maps when new .js files appear in the profile. We'll look for source files in that .js file's directory, and it's parent directories if no source map files are found. This change will be implemented in a follow-up PR to this PR.

We'll need to discuss further if that feature should be on by default, but that should be discussed after we have better performance metrics for when lazy loading is enabled.


aalexand:
@nolanmar511 Great it's coming! Some notes:

CI fails, PTAL?

Also, how about adding top-level description for the PR to mention the motivation, the approach and maybe add screenshots of the resulting profiles?

ts/src/sourcemapper.ts
```diff
@@ -0,0 +1,213 @@
...
+ *  SourceMapConsumer objects that are used to calculate mapping information
+ * @param {string} mapPath The path to the sourcemap file to process.  The
+ *  path should be relative to the process's current working directory
+ * @private
+ */
+async function processSourcemap(
+    infoMap: Map<string, MapInfoCompiled>, mapPath: string) {
+  // this handles the case when the path is undefined, null, or
+  // the empty string
+  if (!mapPath || !_.endsWith(mapPath, MAP_EXT)) {
```
ofrobots: Any reason not to use `mapPath.endsWith(MAP_EXT)`? `String.prototype.endsWith` is [available](https://node.green/) in recent Node 6 and we require a minimum of 6.12.x.

You can get rid of the `lodash` dependency.

ts/src/config.ts
```diff
@@ -149,6 +153,8 @@ export interface ProfilerConfig extends GoogleAuthOptions {
   localProfilingPeriodMillis: number;
   localLogPeriodMillis: number;
   localTimeDurationMillis: number;
+  sourcemapPaths?: string[];
+  sourcemap?: SourceMapper;
```
ofrobots: Why is this object in the config? Previously the config only contained plain data. This is the first object with behaviours that we are adding.

ts/src/config.ts
```diff
@@ -126,6 +127,9 @@ export interface Config extends GoogleAuthOptions {
   // Duration of time profiles collected when using startLocal().
   // For testing with startLocal() only.
   localTimeDurationMillis?: number;
+
+  // Array of paths to files containing source maps.
+  sourcemapPaths?: string[];
```
ofrobots: Does the user have to provide this for source maps to work? On the debug agent, we auto-scan the directory to find source maps. Could we do something similar here?

ts/src/sourcemapper.ts
```diff
@@ -0,0 +1,213 @@
+/**
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
ofrobots: Since this file is largely copied from the Debug agent, please keep the copyright message identical. More specifically, do not adjust the year.

Does it make sense to spin this code into a standalone module? /cc @DominicKramer

ts/src/sourcemapper.ts
nolanmar511: Done.

ts/src/config.ts
nolanmar511: Moved this to the Profiler.
I'd originally put it into the config since the create() method is asynchronous, so I can't call the method from the constructor.  I'm now initializing the source map in the runLoop() function in the profiler.

ts/src/sourcemapper.ts
nolanmar511: Shifted this to third_party.

aalexand:
@nolanmar511 CI is failing?

ts/src/config.ts
```diff
@@ -125,6 +126,18 @@ export interface Config extends GoogleAuthOptions {
   // Duration of time profiles collected when using startLocal().
   // For testing with startLocal() only.
   localTimeDurationMillis?: number;
+
+  // When true, source map support will be disabled.
+  disableSourcemaps?: boolean;
```
aalexand: Nit: `disableSourceMaps`? "source maps" are two words, so I'd expect "maps" to be capitalized. Or are you following the naming convention used by the debugger modules?

aalexand:
@nolanmar511 Is this PR still a draft, or should the title and description be updated?

nolanmar511:
PTAL

aalexand:
@nolanmar511 Sorry, still haven't looked at this PR. Now it says ts/test/test-init-config.ts is in conflict, PTAL?

@ofrobots Could you take a look at this PR too?

aalexand:
Oh, says package.json is also in conflict.

nolanmar511:
cc/ @DominicKramer -- Would you be able to look at source-mapper.ts? It's a modified version of cloud-debug-nodejs's source-mapper (it maps from generated to source, instead of the other way), and your comments there would be helpful.

package.json
```diff
@@ -28,22 +28,29 @@
   "dependencies": {
     "@google-cloud/common": "^0.26.0",
     "@types/console-log-level": "^1.4.0",
+    "@types/lodash": "^4.14.115",
```
ofrobots: It seems that these `@types` dependencies have sneaked in here. Should be in `devDependencies`.

package.json
```diff
@@ -67,7 +74,7 @@
   },
   "files": [
     "out/src",
-    "out/third_party/types",
+    "out/third_party/cloud-debug-nodejs",
```
ofrobots: What is this for?

ts/src/config.ts
```diff
@@ -115,16 +115,25 @@ export interface Config extends GoogleAuthOptions {
...
   // localLogPeriodMillis milliseconds.
   // For testing with startLocal() only.
   localLogPeriodMillis?: number;
 
-
   // Duration of time profiles collected when using startLocal().
   // For testing with startLocal() only.
   localTimeDurationMillis?: number;
+
+  // When true, source map support will be disabled.
```
ofrobots: nit: it would be good to explain what the user visible effects of disabling source maps support are.

ts/src/index.ts
```diff
@@ -122,6 +124,16 @@ async function initConfigMetadata(config: ProfilerConfig):
       config.instance = instance;
     }
   }
+  if (!config.disableSourceMaps) {
+    const mapFiles = await getMapFiles(false, config.workingDirectory);
+    if (config.sourcemapPaths) {
+      mapFiles.concat(config.sourcemapPaths);
+    }
+    mapFiles.forEach((element, idx) => {
```
ofrobots: Would this work:

```js
mapFiles.map(element => path.join(config.workingDirectory, element);
```

ts/src/config.ts
ofrobots: It would good to document the scenarios where users may want to provide this config option. It would be good to document the default behaviour.

ts/src/profiler.ts
```diff
@@ -28,6 +28,7 @@ import {ProfilerConfig} from './config';
 import {logLevelToName} from './index';
 import * as heapProfiler from './profilers/heap-profiler';
 import {TimeProfiler} from './profilers/time-profiler';
+import {create, SourceMapper} from './sourcemapper/sourcemapper';
```
ofrobots: It would be self-documenting to do `import {create as createSourceMapper, SourceMap} ...`.

package.json
nolanmar511: scanner.ts from the debug agent is copied into (and not modified) ts/third_party/cloud-debug-nodejs

ts/src/profiler.ts

package.json
nolanmar511: Done. 

Some of these may have to be added back to be able to the module with typescript, though (the e2e test will fail if this is the case)

ts/src/config.ts

ts/src/config.ts

ts/src/index.ts

package.json
nolanmar511: the `@types/` dependencies that I've kept are necessary for typescript complication.

nolanmar511:
PTAL

package.json
ofrobots: `@types/` dependencies should never be in `dependencies`. Anything in your dependencies will be deployed to users' production. I left some more details thoughts on this topic [elsewhere](https://github.com/grpc/grpc-node/pull/615#discussion_r233265995).

This discussion can probably be moved to a new issue, but I would like to see what issues each of `@types/semver` and `@types/console-log-level` cause. Can you include the e2e compile error message in those cases in a new issue?

package.json
```diff
@@ -33,17 +33,23 @@
     "console-log-level": "^1.4.0",
     "delay": "^4.0.1",
     "extend": "^3.0.1",
+    "findit2": "^2.2.3",
     "gcp-metadata": "^0.9.0",
+    "lodash": "^4.17.10",
```
JustinBeckwith: We've found the full scale lodash package to be really, really slow to load.  I'll keep reading, but chances are you could install sub-packages that are far more targeted.

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
```diff
@@ -0,0 +1,264 @@
...
+  /**
+   * Used to get the file scan results for only the files
+   * whose filenames match the specified regex.
+   *
+   * @param {regex} regex The regex that tests a filename
+   *  to determine if the scan results for that filename
+   *  should be included in the returned results.
+   */
+  selectStats(regex: RegExp): ScanStats|{} {
+    return _.pickBy(this.stats, (ignore, key) => {
```
JustinBeckwith: If this is the only thing you're bringing in lodash for, you can do this a little less elegantly with a reduce:

```js
Object.keys(object).reduce((acc, key) => {
  if (regex.test(key)) {
    acc[key] = object[key];    
  }
  return acc;
}, {});
```

ts/src/sourcemapper/sourcemapper.ts
```diff
@@ -0,0 +1,230 @@
...
+      column: (pos as {} as {col: number}).col
+    };
+  }
+}
+
+export async function create(sourcemapPaths: Map<string, string>):
+    Promise<SourceMapper> {
+  const limit = pLimit(CONCURRENCY);
+  const mapper = new SourceMapper();
+  const promises: Array<Promise<void>> = [];
```
JustinBeckwith: This pattern could be rewritten as a map:
```js
const promises = sourcemapPaths.map((parentDir, path) => {
   return processSourcemap(mapper.infoMap, path, parentDir);
});
```

Feel like I needed to make up for the extra few lines I caused with my lodash feedback 😆

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
JustinBeckwith: It's also worth asking - if you expect to manage the object this way, should it be a `Map`? (honestly asking, not sure)

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
```diff
@@ -0,0 +1,264 @@
...
+ * based on the contents.
+ *
+ * @param {!Array<string>} fileList array of filenames
+ * @param {Boolean} shouldHash whether a hash should be computed
+ * @param {!function(?Error, ?string, Object)} callback error-back style callback
+ *    returning the hash-code and an object containing file statistics.
+ */
+// TODO: Typescript: Fix the docs associated with this function to match the
+// call signature
+function computeStats(
```
JustinBeckwith: Can you just make this an async function, and drop the promise wrapping?

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
JustinBeckwith: Aye, I have a PR to remove lodash from that as well.  That's the problem with lodash, you let it in for one method, then it lets in all of it's friends :)

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
nolanmar511: This file is copied without modifications from https://github.com/googleapis/cloud-debug-nodejs.
cc/ @DominicKramer

ts/src/sourcemapper/sourcemapper.ts
nolanmar511: I get a typescript error when I try this:
```
Property 'map' does not exist on type 'Map<string, string>'.
```

nolanmar511:
PTAL

ts/src/sourcemapper/sourcemapper.ts
JustinBeckwith: eh, either way, this is fine :)

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
JustinBeckwith: This still needs to be addressed.  We changed debug to drop the full lodash dependency and instead pull in `lodash.pickby`.  That or the reduce pattern above should work fine.

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
JustinBeckwith: Even if copied, I still think we should do this correctly, no?

nolanmar511: I think this is best addressed in cloud-debug-nodejs. 
I have this file copied in as a subtree. To keep this maintainable, I really don't want to add changes to this.

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
nolanmar511: This is now done.
I got the changes from cloud-profiler-nodejs.

ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
JustinBeckwith: ahhhh, got it!

ofrobots:
Is this ready for review, or should I wait for a rebase once #345 has landed?

nolanmar511:
It probably makes sense to wait to review until #345, since I'm hoping this change will be a bit easier to review once #345 has landed.

nolanmar511:
#345 has landed. 
PTAL

ts/src/config.ts

ts/src/sourcemapper/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
aalexand: I think this license and the whole sourcemapper directory shouldn't appear in this PR since it's already submitted?

package.json
```diff
@@ -71,12 +75,13 @@
...
     "sinon": "^7.0.0",
     "source-map-support": "^0.5.6",
+    "tmp": "0.0.33",
     "ts-mockito": "^2.2.5",
     "typescript": "~3.2.0"
   },
   "files": [
     "out/src",
-    "out/third_party/types",
+    "out/third_party/cloud-debug-nodejs",
```
aalexand: The "cloud-debug-nodejs" part doesn't look right as we don't have that third_party dep anymore?

ts/src/sourcemapper/LICENSE
nolanmar511: I forgot to remove the LICENSE and README after rebasing. 

sourcemapper.ts will show changes to reverse the mapping,

package.json
nolanmar511: There's still a third_party dependency. 
scanner.ts from cloud-debug-nodejs (which we still use in the source mapper) was copied without modification (it's a subtree), so it's in third_party.

aalexand:
@nolanmar511 Probably missing the obvious, but I don't see cloud-debugger-nodejs in https://github.com/googleapis/cloud-profiler-nodejs/tree/master/third_party? Oh, it's in https://github.com/googleapis/cloud-profiler-nodejs/tree/master/ts/third_party/cloud-debug-nodejs? Is there a reason it's in a different location?

Also, given that it's 200 lines of code, would it make sense to just copy that source file over rather than have a dependency?

nolanmar511:
As far as why it's in a different location, it's typescript code, so needs to be within the "ts" location to be compiled.

I wouldn't have strong opinions against moving it, but this PR is busy enough that if we decide it should be moved, could we file an issue and move it after this PR is merged?

ts/src/config.ts
```diff
@@ -115,16 +115,37 @@ export interface Config extends GoogleAuthOptions {
...
+  // Map of files containing source maps to directories where generated code
+  // refered to by source maps is located. Source maps can be used to
+  // map javascript locations in transpiled code that is running to original
+  // locations in the source code.
+  // All .js.map files which are available in the runtime and within the
+  // application's working directory are added to this map. It is assumed that
+  // these maps will be in the same directory as the generated code.
+  // This can be used to specify source maps which are available in the runtime,
+  // but outside of the working directory, or not in the same directory as the
+  // generated code.
```
ofrobots: After reading this a few times, I am still confused about what this does. As a user, what are the scenarios where I would need to specify this config option? How do I know if the default value would just work for my project.

aalexand: +1

ts/src/config.ts
```diff
@@ -115,16 +115,31 @@ export interface Config extends GoogleAuthOptions {
...
 
-
   // Duration of time profiles collected when using startLocal().
   // For testing with startLocal() only.
   localTimeDurationMillis?: number;
+
+  // When true, source map support will be disabled.
+  // All locations in the profiler will reference locations in the running
+  // JavaScript.
+  disableSourceMaps?: boolean;
```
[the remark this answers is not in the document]
nolanmar511: Switching to sourceMap everywhere to match names in the "source-map" module.

ts/src/config.ts
```diff
@@ -115,16 +115,31 @@ export interface Config extends GoogleAuthOptions {
...
+
+  // All .js.map files which are available in the runtime and within the
```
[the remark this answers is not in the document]

testing/integration_test.go
```diff
@@ -199,6 +200,11 @@ func TestAgentIntegration(t *testing.T) {
 		ComputeService: computeService,
 	}
 
+	wantProfiles := []profileSummary{
```
aalexand: The current checks appear to only verify that there is any sample for busyLoop and any sample for busybench.ts. It would be good to have a more strict check that there are samples in busyLoop function and busybench.ts source file.

ts/src/config.ts
```diff
@@ -115,16 +115,36 @@ export interface Config extends GoogleAuthOptions {
...
+  // process.cwd() and the source maps of dependencies can normally be found
+  // via recursive search in process.cwd()/node_modules so the default value
+  // should work well pretty much always.
+  sourceMapSearchPath?: string[];
+
+  // Generally should not be set.
+  // Can be used to add a source map not within a directory any of the
+  // directories specified by sourceMapSearchPath, by specifying the absolute
+  // path to the source map in this array.
+  sourceMapPaths?: string[];
```
aalexand: My previous comment suggested (probably in a way that was not clear enough) that this configuration option is dropped. Is there a reason you decided to keep it? I am fine to keep it if you think there is a usage model for it that is important for most Node.js developers, but then clarify that in the doc string - currently the documentation for this option is obscure in terms of why it's important to have this option and is more of "nice to have though we don't really know why and when" knob.

ts/src/index.ts
```diff
@@ -84,6 +87,12 @@ function initConfigLocal(config: Config): ProfilerConfig {
   const mergedConfig =
       extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
+  // if sourceMapSearchPath is empty array, extend will override its value
```
aalexand: Nit: "if" -> "If" (keep comments proper sentences)

ts/src/sourcemapper/sourcemapper.ts
```diff
@@ -185,56 +173,54 @@ export class SourceMapper {
...
-                                 // zero-based output.
-      column: mappedPos.column   // SourceMapConsumer uses
-                                 // zero-based column
-                                 // numbers which is the
-                                 // same as the expected
-                                 // output
+      file: path.resolve(entry.mapFileDir, pos.source),
+      line: pos.line || undefined,
+      name: pos.name || location.name,
+      // TODO: The `sourceMap.Position` type definition has a `column`
```
aalexand: Just curious: this TODO seems like something we can check / address as part of this PR. Why it's deferred?

ts/src/config.ts
nolanmar511: Updated this so this field is only in ProfilerConfig (the config passed to the Profiler class), not in Config (the config a user passes to the `start` function).

ts/src/index.ts

testing/integration_test.go
nolanmar511: This is WIP.
Adding a function to support this to proftest.go (https://code-review.googlesource.com/c/gocloud/+/36950)

ts/system-test/test-start.ts
```diff
@@ -86,7 +86,7 @@ before(async () => {
 
   // start profiling and wait to collect profiles.
   const profiler = require('../src/index');
-  profiler.start({credentials: fakeCredentials});
+  profiler.start({credentials: fakeCredentials, logLevel: 5});
```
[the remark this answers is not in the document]
nolanmar511: This was added for debugging (a log level of 4 should be sufficient -- putting 5 is a bad habit of mine that I should work to break).

ts/src/profiler.ts
```diff
@@ -251,6 +252,7 @@ export class Profiler extends ServiceObject {
   private deployment: Deployment;
   private profileTypes: string[];
   private retryer: Retryer;
+  private sourceMap: SourceMapper|undefined;
```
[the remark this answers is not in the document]
nolanmar511: Done!

ts/src/config.ts
```diff
@@ -172,5 +188,7 @@ export const defaultConfig = {
 
   localProfilingPeriodMillis: 1000,
   localLogPeriodMillis: 10000,
-  localTimeDurationMillis: 1000
+  localTimeDurationMillis: 1000,
+  sourceMapSearchPath: [process.cwd()],
+  sourceMapPaths: [],
```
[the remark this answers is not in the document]
nolanmar511: Right now, this is for convenience. It's nice to be able to return a ProfilerConfig from initConfigLocal, and convenient to use the defaultConfig to add the sourceMapPaths field.

However, I think one reason for your comment is that defaultConfig should be able to be used as an object of type Config. With the sourceMapsPaths field, that wouldn't be the case right now. 

This is WIP as I think about how to do initialization better.

ts/src/config.ts
```diff
@@ -148,6 +162,8 @@ export interface ProfilerConfig extends GoogleAuthOptions {
   localProfilingPeriodMillis: number;
   localLogPeriodMillis: number;
   localTimeDurationMillis: number;
+  sourceMapSearchPath: string[];
```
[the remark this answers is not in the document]
nolanmar511: This is WIP.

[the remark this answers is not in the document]
aalexand: Can you just delete the sourceMapPaths field from ProfilerConfig? Or it won't build then?

ts/src/config.ts
```diff
@@ -115,16 +115,30 @@ export interface Config extends GoogleAuthOptions {
...
+  // or dependency written in TypeScript: the source file paths that
+  // the profiler observes in the profiling data are the transpiled *.js files;
+  // to attribute the data back to the TypeScript source that the developer
+  // wrote source map needs to be generated, distribute and used.
+  //
+  // The source map of the application typically resides directly in
+  // process.cwd() and the source maps of dependencies can normally be found
+  // via recursive search in process.cwd()/node_modules so the default value
+  // should work well pretty much always.
+  sourceMapSearchPath?: string[];
```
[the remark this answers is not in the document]
nolanmar511: I think I misunderstood a previous comment. 

Added this option back.

ts/src/config.ts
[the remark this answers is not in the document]
aalexand: It doesn't build how? I.e. what are the errors?

Making the field optional still feels like a massage rather than a fix. Conceptually, you have a thing A here that is used to produce thing B as part of initialization and the thing B is what you feed into the profiler class. Still carrying thing A as part of that input to the profiler is redundant / confusing.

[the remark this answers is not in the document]
aalexand: I don't want to trigger a major refactoring though, so I mostly wonder if there is an easy / elegant way to address this.

ts/src/config.ts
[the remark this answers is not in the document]
aalexand: sourceMapSearchPath shouldn't be part of ProfilerConfig either though, as we discuss in a parallel thread.

ts/src/config.ts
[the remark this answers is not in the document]
nolanmar511: I don't think there's a really simple way to change this. Here are the solutions I can think of. All would require a bit of refactoring of the current code, but I don't think any would be really major.

Option 1:

`initLocalConfig` could return a different type (which I'd call `LocalConfig`, possibly). 

Option 1A:
`ProfilerConfig` would have all current fields (fields present before any source map support), plus the following:
```
export interface ProfilerConfig extends GoogleAuthOptions {
  ...
  sourceMapPath: string[];
  disableSourceMaps: boolean;
}
```

`LocalConfig` would be unexported, and defined in index.ts. `LocalConfig` would have all fields that ProfilerConfig in current source code has, plus the following:
```
interface LocalConfig extends GoogleAuthOptions {
  ...
  sourceMapSearchPath: string[];
  disableSourceMaps: boolean;
}
```

Option 1B:
Roughly the same as 1A, but while we currently initialize the SourceMapper when calling `profiler.start()`, for this option, the initialized SourceMapper would be part of the ProfilerConfig (and the config would be initialized in `initConfigAsync`). 

`ProfilerConfig` would have all current fields, plus the following:
```
export interface ProfilerConfig extends GoogleAuthOptions {
  ...
  sourceMapper?: SourceMapper;
}
```
Source map support would be disabled when the sourceMapper is undefined.

`LocalConfig` would be  the same as in 1A.


Option 2:
Right now, we find source maps files during config initialization, and initialize the SourceMapper when we call `profiler.start()`.

We could shift finding source map files given `sourceMapSearchPath` to when we call `profiler.start()`

Then `ProfilerConfig` would have all current fields (fields present before any source map support), plus the following:
```
export interface ProfilerConfig extends GoogleAuthOptions {
  ...
  sourceMapSearchPath: string[];
  disableSourceMaps: boolean;
}
```

If none of these options are good or you think a hybrid approach might be better, let me know, and I can move this to a design document --I think the structure of a design document might be better for discussions where we're considering multiple options.

I have some preference for options 1B and 2, because with those options group all of the work done to initialize the SourceMapper into one place. With either of these options, we could have `sourceMapSearchPath`  be an argument for `create` (in sourcemapper.ts) instead of `sourceMapPaths`. 

I think I could implement option 2 with minimal changes, so I'll go ahead and do that.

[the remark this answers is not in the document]
nolanmar511: Implemented option 2.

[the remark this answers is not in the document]
nolanmar511: Implemented option 2.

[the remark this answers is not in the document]
aalexand: Option 2 SGTM as far as I can tell from the description. There might be some wrinkles in terms of it being asynchronous and error checking but hopefully not. Let's see if we can settle the change here, a design doc may be an overkill.

testing/integration_test.go
aalexand: I also don't mind this enhancement to come later.

ts/src/config.ts
```diff
@@ -115,16 +115,35 @@ export interface Config extends GoogleAuthOptions {
...
+
+  // When true, source map support will be disabled.
+  // All locations in profiles will reference locations in the running
+  // javascript.
```
aalexand: Nit: javascript -> JavaScript or JS.

ts/src/config.ts
```diff
@@ -148,6 +167,8 @@ export interface ProfilerConfig extends GoogleAuthOptions {
   localProfilingPeriodMillis: number;
   localLogPeriodMillis: number;
   localTimeDurationMillis: number;
+  sourceMapSearchPath: string[];
+  disableSourceMaps?: boolean;
```
aalexand: Is it intended to be optional?

ts/src/index.ts
```diff
@@ -84,6 +84,13 @@ function initConfigLocal(config: Config): ProfilerConfig {
   const mergedConfig =
       extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
+  // If sourceMapSearchPath is empty array, extend will override its value
```
aalexand: This comment could be clarified a bit - it's not clear what's wrong with the default value from mergedConfig.

ts/src/sourcemapper/sourcemapper.ts
```diff
@@ -185,56 +173,68 @@ export class SourceMapper {
...
+      file: path.resolve(entry.mapFileDir, pos.source),
+      line: pos.line || undefined,
+      name: pos.name || location.name,
+      column: pos.column,
     };
   }
 }
 
-export async function create(sourcemapPaths: string[]): Promise<SourceMapper> {
+export async function create(sourceMapSearchPath: string[]):
```
aalexand: Nit: consider dropping "sourceMap" prefix from names where it's clear from the context. E.g. sourceMapSearchPath can be shortened to searchPath without affecting readability I think, similar for some local vars below.

ts/src/sourcemapper/sourcemapper.ts
```diff
@@ -185,56 +173,68 @@ export class SourceMapper {
...
   } catch (err) {
     throw new Error(
-        'An error occurred while processing the sourcemap files' + err);
+        'An error occurred while processing the source map files' + err);
   }
   return mapper;
 }
+
+async function getMapFiles(
+    shouldHash: boolean, baseDir: string): Promise<string[]> {
```
aalexand: `shouldHash` argument doesn't seem to be used?

ts/src/sourcemapper/sourcemapper.ts
aalexand: I might have asked this before, but not positive: how long does the search of map files take for a reasonably sized Node.js app?

ts/src/config.ts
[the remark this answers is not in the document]
nolanmar511: Marking this as resolved for now -- we've settled on a general approach and the approach is now implemented. So we can make more specific comments about that approach in the code.

ts/src/config.ts

ts/src/config.ts
nolanmar511: Sorry -- copy-paste error.

ts/src/index.ts
nolanmar511: I tried to clarify the comment more. 

Basically, extend seems to treat a field set with an empty array in the same way it would treat an unset or undefined field -- so this is needed.

ts/src/sourcemapper/sourcemapper.ts
nolanmar511: I left the "map" prefix in some places (since this file imports "path", I used "mapPath" in places; and also used "mapFiles" and "mapPath" for consistency).

nolanmar511:
Right now, we don't look for source maps in node_modules. 
Should I patch scanner.ts to look for map files in node_modules?

ts/src/sourcemapper/sourcemapper.ts
nolanmar511: I modified scanner.ts so it would search node_modules. With that modification, there were 1264 map files, and it took 0.75s for create() to run on my corp desktop.

nolanmar511:
PTAL

aalexand:
@nolanmar511 The PR description should be updated? It says "Wait to review until #345 is merged" but that PR is merged already. The rest of the description does not appear to holistically describe the added feature either, it should read as a short mini-story about the added feature.

testing/integration_test.go
aalexand: @nolanmar511 This is done?

testing/integration_test.go
```diff
@@ -293,6 +299,12 @@ func TestAgentIntegration(t *testing.T) {
 					t.Errorf("QueryProfiles(%s, %s, %s, %s, %s) got error: %v", tc.ProjectID, tc.name, startTime, endTime, wantProfile.profileType, err)
 					continue
 				}
+				if wantProfile.sourceFile != "" {
+					if err := pr.HasFunctionInFile(wantProfile.functionName, wantProfile.sourceFile); err != nil {
+						t.Errorf("Function %s not found source file %s in profiles of type %s: %v", wantProfile.functionName, wantProfile.sourceFile, wantProfile.profileType, err)
```
aalexand: "Function %s not found source file %s" doesn't read well.

ts/src/index.ts
```diff
@@ -84,6 +84,15 @@ function initConfigLocal(config: Config): ProfilerConfig {
   const mergedConfig =
       extend(true, {}, defaultConfig, envSetConfig, envConfig, config);
 
+  // If sourceMapSearchPath is empty array in config, extend will override
+  // its value with the default value in mergedConfig (as if
+  // sourceMapSearchPath were not specified in config), rather than keep the
+  // empty array as sourceMapSearchPath's value in mergedConfig.
+  if (Array.isArray(config.sourceMapSearchPath) &&
```
aalexand: I see it now, thanks for the updated comment. So it's basically about distinguishing undefined sourceMapSearchPath vs. empty array sourceMapSearchPath values and hacking things up so that the latter can be used as an override rather then be ignored as in the current extend() logic.

What's the use case for specifying empty sourceMapSearchPath array though? Is this to disable source maps support? But there is disableSourceMaps option for that. Having two ways for that seems redundant, if that's the intention.

ts/src/profiler.ts
```diff
@@ -317,7 +319,17 @@ export class Profiler extends ServiceObject {
    * on the type of profile to be collected, this problem will be logged at the
    * error level and getting profile type will be retried.
    */
-  start() {
+  async start() {
```
aalexand: For my eduction: if this function async now, does it mean it implicitly returns a promise?

testing/integration_test.go
nolanmar511: Yes! This is done. 

I lost this comment, sorry.

ts/src/index.ts
nolanmar511: Yeah -- it would be redundant. 
`disableSourceMaps=true` is better (certain functions won't be called).
But I also hate the idea that a user could pass in an empty array and it would be replaced with the default value, and source maps would be found and used. 

So I'm slightly for keeping this, but it's not a strong feeling.

ts/src/profiler.ts
nolanmar511: Yes, this implicitly returns a `Promise<void>`.
I marked the return type.

ts/src/index.ts
ofrobots: I think you should throw an error if the user passes an empty `sourceMapSearchPath` and advise users to use `disableSourceMaps` instead.

aalexand:
@nolanmar511 I agree with @ofrobots that it would make sense to throw an error if the user passes an empty sourceMapSearchPath and advise users to use disableSourceMaps instead. Having two ways to do something is a bit confusing and piles up more code for maintenance and testing.

aalexand:
> excluding any directories named node_modules or .git, and find the source maps within the working directory (or other specified directories).

@nolanmar511 Like @ofrobots mentioned elsewhere, it's not clear why ".git" is special here. Why did you decide to include this directory here? If it really needs to be included, what about other version control systems like Mercurial?

Could you also include some numbers about the performance of the recursive scan in the PR description? You mentioned some numbers before, but I think those were measured on your workstation and I don't recall what was the app. Please include numbers for something closer to what users usually run in the cloud - e.g. some simple express package based application running on n1-standard-1 GCE VM with usual, not SSD, PD. Typical workstation today uses local SSD and has many cores, which is very much different from IOPS-throttled and single core GCE VM environment.

ts/src/sourcemapper/sourcemapper.ts
aalexand: See my other comment on the performance measurement. In short, use an express module based app (as it's not clear which app you used here and express-based app is a good decent size app) and measure on an n1-standard-1 GCE VM with ordinary (not SSD) disk. Please add the numbers of your measurement to the PR description.

ts/src/index.ts
nolanmar511: Done!
That's a much better solution.

nolanmar511:
".git" and "node_modules" are both special, because neither are currently explored for source maps. scanner.ts skips these files: https://github.com/googleapis/cloud-profiler-nodejs/blob/6a5d74d09955d3434f5063aa6c5be088cfe7f82f/ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts#L197-L199

And working on performance numbers on n1-standard GCE VM.

nolanmar511:
@aalexand  -- PTAL, I've added a few experiments and next steps to the PR description.

ts/src/config.ts
```diff
@@ -115,16 +115,35 @@ export interface Config extends GoogleAuthOptions {
...
   // Duration of time profiles collected when using startLocal().
   // For testing with startLocal() only.
   localTimeDurationMillis?: number;
+
+  // List of directories recursively searched for *.js.map files. Defaults to
+  // process.cwd().
+  //
+  // The profiler uses these files to re-map the source file paths in the
+  // profiles. The most common use case of having a source map is an application
+  // or dependency written in TypeScript: the source file paths that
```
aalexand: The text here says "or dependency", but that's not correct since node_modules is not searched yet?

aalexand:
> Long term, we would like to make it possible for users to automatically use source maps from node_modules. ... We'll need to discuss further if that feature should be on by default.

I am not sure it should be long term, I think we should implement it as a follow-up change. And with the lazy processing of source maps it should be fine to enable that by default I think.

ts/src/config.ts

PR #271, opened 2018-08-09
chore: use console-log-level for logging

nolanmar511:
This is necessary to upgrade `@google-cloud/common` to 0.21.0.

Logging was a part of common previously, but has been moved (more details on https://github.com/googleapis/nodejs-common/issues/204)

### DIFF
--- a/ts/src/index.ts
+++ b/ts/src/index.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import {Logger, util} from '@google-cloud/common';
+import {util} from '@google-cloud/common';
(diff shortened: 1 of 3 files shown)


ts/src/index.ts
```diff
@@ -188,9 +189,26 @@ export async function start(config: Config = {}): Promise<void> {
   profiler.start();
 }
 
+const LEVELNAMES: consoleLogLevel.LogLevelNames[] =
```
aalexand: Should this be LEVEL_NAMES?

nolanmar511:
Merging despite drop in code coverage.

PR #274, opened 2018-08-17
chore: delete unused dependencies

fhinkel:


fhinkel:
I don't have write access, so if somebody else can merge this for me :bowing_woman:

PR #279, opened 2018-08-21
chore(package): upgrade to delay@3.1.0

ofrobots:
The package started shipping types as part of the official package.
These are breaking changes w.r.t previous `@types/delay` types. We need
to import the module differently now for things to work correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@google-cloud/common": "^0.20.3",
     "bindings": "^1.2.1",
     "console-log-level": "^1.4.0",
-    "delay": "~3.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "@google-cloud/common": "^0.20.3",
     "bindings": "^1.2.1",
     "console-log-level": "^1.4.0",
-    "delay": "~3.0.0",
+    "delay": "^3.1.0",
```
DominicKramer: Shouldn't `@types/delay` also be removed?

ofrobots: 👍

PR #283, opened 2018-08-23
feat: use smaller HTTP dependency

fhinkel:
Use teeny-request instead of large `request` dependency. According to
[require-so-slow](https://www.npmjs.com/package/require-so-slow),
load time improves from 1500 ms to 600 ms.

Before with `request`:
screen shot 2018-08-23 at 4 04 23 pm

After with `teeny-request`:
screen shot 2018-08-23 at 4 04 11 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
     "pify": "^4.0.0",
     "pretty-ms": "^3.1.0",
     "protobufjs": "~6.8.6",
-    "request": "^2.88.0",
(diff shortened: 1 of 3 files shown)


fhinkel:
The failing tests are broken for me without the change, too. Need to figure out how to debug this locally.

aalexand:
This looks good to me, but would like @nolanmar511 to LGTM as well.

PR #311, opened 2018-10-18
feat: validate service name before profiler creation

### LINKED ISSUE #309: Add client-side validation of of service name
The profile server requires that the service name match a particular regex (`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`).

Right now, we only validate this on the server-side, but we should also validate this in the profiling agent, which would allow us to give users clearer error messages.

nolanmar511:
fixes #309 

Without this fix, profiler continue to try to create profiles, logging failures to create the profile at the "warn" level.

### DIFF
--- a/ts/test/test-init-config.ts
+++ b/ts/test/test-init-config.ts
@@ -204,16 +204,34 @@ describe('createProfiler', () => {
       disableHeap: true,
       disableTime: true,
     };
-    createProfiler(config)
-        .then(() => {
-          assert.fail('expected error because no service in config');
-        })
-        .catch((e: Error) => {
-          assert.strictEqual(
-              e.message,
-              'Could not start profiler: Error: Service must be specified in the configuration');
-        });
+    try {
+      await createProfiler(config);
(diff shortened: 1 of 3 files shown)


ts/test/test-init-config.ts
```diff
@@ -215,6 +215,27 @@ describe('createProfiler', () => {
...
+         disableTime: true,
+       };
+       createProfiler(config)
+           .then(() => {
+             assert.fail('expected error because no service in config');
+           })
+           .catch((e: Error) => {
+             assert.strictEqual(
+                 e.message,
+                 'Could not start profiler: Error: Service must be specified in the configuration');
```
aalexand: It would probably be good if the error here would tell the user that the service was specified but didn't match the required pattern rather than say that the service is not specified?

nolanmar511: Fixed -- turned out to be a copy-paste error, paired with a problem in how I'd written the test.

PR #323, opened 2018-10-30
fix: fix continuous e2e tests

nolanmar511:


aalexand:
@nolanmar511 Please merge once the test passes.

package.json
```diff
@@ -63,7 +63,7 @@
     "sinon": "^7.0.0",
     "source-map-support": "^0.5.6",
     "ts-mockito": "^2.2.5",
-    "typescript": "~3.1.0"
+    "typescript": "^3.1.4"
```
JustinBeckwith: 👋 That's generally a scary thing to do.  TypeScript likes to make compiler changes between semver minor releases, which can cause your program to just stop compiling :(  It's obviously up to you how you want to manage this, but using a `^` range specifier will cause breaks on master at some point for you.

nolanmar511: This may not be necessary -- the compilation failed on our benchmark. 

Just trying everything to get the benchmark working, and I'll go back from there.

nolanmar511: Done

PR #324, opened 2018-10-31
e2e test should finish when there is an error

nolanmar511:


aalexand:
@nolanmar511 Please merge once the test finishes.

Issue #336, opened 2018-11-20
What to do about @types dependencies that aren't in devDependencies


nolanmar511:
Typescript modules will not compile without @types/console-log-level and @types/semver modules:
https://sponge.corp.google.com/target?id=3c7a9b78-a7be-45f7-be4b-80b7cc99afa1&target=cloud-profiler/nodejs-agent/presubmit

JustinBeckwith:
There are really two ways to go about this:
1. Don't expose other modules interfaces publicly
2. Ship the `@types` package in the `dependencies` section

1 is ideal, sometimes 2 happens 🤷‍♂️

ofrobots:
IMO, 2 can be avoided in most situations, as here.

PR #337, opened 2018-11-20
fix: do not export console-log-level publicly

ofrobots:
console-log-level types are an implementation detail. Exporting them
meant that users needed to have @types/console-log-level installed to be
able to use this module. This requires either that the users know that
they need @types/console-log-level in their devDependencies or we need
to decalare them in our main dependencies.

Fixes: https://github.com/googleapis/cloud-profiler-nodejs/issues/336


ts/src/logger.ts
```diff
@@ -0,0 +1,39 @@
...
+  } else if (level > 4) {
+    level = 4;
+  }
+  return LEVEL_NAMES[level];
+}
+
+export function createLogger(level?: number) {
+  return consoleLogLevel(
+      {stderr: true, prefix: pjson.name, level: logLevelToName(level)});
+}
```
nolanmar511: Nit: Missing new line at end.

ofrobots:
```suggestion
}

```

@nolanmar511 note this cool new feature in github that lets you suggest nits as fixes :)

PR #341, opened 2018-11-20
0.2.3

nolanmar511:
Release notes at https://github.com/googleapis/cloud-profiler-nodejs/releases/edit/untagged-899bc7bad2b080f9d8c0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/profiler",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "description": "Adds support for Stackdriver Profiler to node.js applications",
   "repository": "googleapis/cloud-profiler-nodejs",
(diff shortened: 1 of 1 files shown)


nolanmar511:
I feel like moving to 1.0.0 after we finish all issues for the Agent Beta milestone list (https://github.com/googleapis/cloud-profiler-nodejs/milestone/2) might make sense?

ofrobots:
In semver, and specially in the npm ecosystem, `1.0.0` is about API stability rather than support levels. Do we expect API changes in the beta milestone? If not, I would recommend going to `1.0.0` on the next release.

nolanmar511:
The source map support (#268) and accurate line number (#258) changes both are likely to change functions which are exported, though not the `start()` function, which is the only function we intend users to use.

Issue #47, opened 2017-11-03
Publish pre-built binaries for native components


ofrobots:
This is probably not needed for EAP, but eventually we should start using `node-pre-gyp` and shipping precompiled binaries to avoid users having to compile at installation time.

aalexand:
Do we still need this? From our discussion I got impression that Node.js environments typically have the compiler installed since many packages with native components do not ship prebuilt binaries?

ofrobots:
We do need this, but perhaps not imminently.

Most customers would have the prerequisite necessary to build the native components in the deployment environment, but many won't. Even folks who _can_ compile will still see compile time cost incurred at deployment time, making their deployments slow.

Precompiled binaries improves the overall UX, so we should have them (eventually).

evanlucas:
Having to compile native addons at build time slows down our builds quite a bit, so this is something that we would definitely benefit from.

aalexand:
@evanlucas Is the profiler agent your only dependency that has to build the native part and the rest is coming precompiled, or the agent is just making things worse among others? Just curious to which extent shipping prebuilt native components is a standard among Node deps.

evanlucas:
we currently ship 20 services that use native addons and they all are precompiled. Occasionally, one of our dependencies will not offer a prebuild (like when we bump our node version in our base docker image) and it more than triples our build times. We aren't currently shipping with the profile agent at this point, but I am about to start experimenting with it.

aalexand:
@nolanmar511 Looking at modules called prebuild and prebuild-ci should be useful for the task.

nolanmar511:
prebuild-ci seems to require using a personal access token (which would be associated with a github account, rather than a project). Is there a good way around that?

nolanmar511:
https://github.com/grpc/grpc-node builds pre-compiled binaries for node. Prebuilt binaries are built with kokoro and stored in a GCS bucket 
I think following their pattern would work.
That project uses scripts in https://github.com/grpc/grpc-node/tree/master/packages/grpc-native-core/tools/run_tests/artifacts to build. Scripts called by kokoro are in https://github.com/grpc/grpc-node/tree/master/tools/release.

nolanmar511:
While grpc uses kokoro for prebuilding it's binaries, I'm not certain that kokoro is right for our use case. At least, it might not be the right choice if we want to stick with our current release process. 
[...]

PR #356, opened 2018-12-06
chore: start using node-pre-gyp

nolanmar511:
This is preparation for #47.

Right now, we call NODE_MODULE() twice (in sampling-heap-profiler, and in time-profiler), and generate two separate binaries (one for sampling-heap-profiler, and one for time-profiler).

`node-pre-gyp` requires that the "module_name" in the package.json has to match the name passed to the NODE_MODULE macro. There isn't a way to specify the "module_name" twice, this change refactors, so that the NODE_MODULE macro is only called once.


aalexand:
@nolanmar511 What is a "node" file?

It is also not clear why you had to switch to use objects instead of free functions, could you elaborate on that (here or just in the PR description).

nolanmar511:
Tried to clarify the description, and switched back to using free functions.

aalexand:
@nolanmar511 Would things be simpler overall if we'd just put all bindings to a single profiler.cc file? The interface to both of time and heap profiler is tiny and I feel that we have unnecessary complexity when we put them into separate files and then have two Init functions and a wrapper around that to call both Init's etc.

nolanmar511:
I'm not opposed to putting everything one one file. There are three reasons I separated these out.  First, I feel like logically separating these might make it easier to ultimately transition to using the sampling-heap-profiler node module, which is something we've planned to do. Next is that we'd have to rename StartProfiling, StopProfiling, and SetSamplingInterval for the time profiler. And finally, if we add ever add profile types, the single profile.cc file could get too large.

None of these things are particularly strong reasons not to put everything in one file, though.

Another alternative is to keep the structure, but move everything in TimeProfilerInit and SamplingHeapProfilerInit directly into the Init function in profiler.cc.

aalexand:
@nolanmar511 I think I'd put everything in one file but would skip on the renaming part. Yes, the names are slightly inconsistent, but that's minor I think.

PR #368, opened 2018-12-12
DO NOT REVIEW: experimenting with pre-compiled binaries

nolanmar511:


nolanmar511:
Abandoning this release. 
I was experimenting with using a go job, but I think it's similar to use multiple kokoro configs.

PR #376, opened 2018-12-15
fix: add empty go file to allow integration test to be run as go test

nolanmar511:


aalexand:
@nolanmar511 Why did this error started to happen and didn't happen before? Did version of Go get updated in Kokoro VMs?

aalexand:
I think https://github.com/googleapis/cloud-profiler-nodejs/pull/377 is a better solution, I merged it.

nolanmar511:
#377 is better fix. 

Abandoning this change

nolanmar511:
And, we don't print the version of go used by the kokoro machine, so I'm not sure if the go version changed. My assumption, though, is that the go version did change.

PR #377, opened 2018-12-15
Fix "no non-test Go files" error.

aalexand:


testing/integration_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/GoogleCloudPlatform/google-cloud-go/profiler/proftest"
+	"cloud.google.com/go/profiler/proftest"
```
aalexand: This change is not really related to the fix, but using consistent path with https://code.googlesource.com/gocloud/+/master/profiler/integration_test.go#31.

PR #380, opened 2018-12-28
build: check broken links in generated docs

kinwa91:


aalexand:
Some CI appear to fail here.

PR #382, opened 2019-01-07
fix: scanner.ts should recursively search .git and node_modules directories

nolanmar511:
We want modules to also have source map support.

### DIFF
--- a/ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
+++ b/ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
@@ -194,7 +194,7 @@ function findFiles(baseDir: string, regex: RegExp): Promise<string[]> {
 
     find.on('directory', (dir: string, ignore: fs.Stats, stop: () => void) => {
       const base = path.basename(dir);
-      if (base === '.git' || base === 'node_modules') {
+      if (base === '.git') {
         stop();  // do not descend
       }
     });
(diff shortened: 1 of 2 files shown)


ts/third_party/cloud-debug-nodejs/src/agent/io/scanner.ts
```diff
@@ -194,9 +194,6 @@ function findFiles(baseDir: string, regex: RegExp): Promise<string[]> {
 
     find.on('directory', (dir: string, ignore: fs.Stats, stop: () => void) => {
       const base = path.basename(dir);
-      if (base === '.git' || base === 'node_modules') {
```
ofrobots: I can understand the motivation for `node_modules`, but why descend into .git?

nolanmar511:
Yeah -- I'm a bit worried about the possible memory consumption of this. I tried on cloud-profiler-nodejs, and there were ~1600 .js.map files (if I remember correctly). So that would be a lower bound on the number of map files. 

Maybe we could add a flag to enable searching node_modules (so not look at dependencies source maps by default). Or maybe even allow users to specify the names of modules which should be searched for source maps. 

I'm envisioning a use case where someone runs code which is maybe a thin wrapper around a few modules that they have, and they want to have source maps for those modules included.

ofrobots:
Yeah, a flag would make sense, but I wonder whether the default should be to scan the `node_modules` directory by default or not. It depends on whether the dependencies were authored by 3rd party or 1st party as you mention in your example. Another use-case would be if a (3p) library is shipped as a bundle (single file). Based on this I am leaning towards default to use source maps for files even in `node_modules` for the profiler.

 Specifying individual modules piecemeal seems very burdensome, so I would advise against that option.

Regardless of the default, I think we should be efficient in scanning for source maps. WDYT about what I proposed in my previous comment?

nolanmar511:
Closing this PR -- in meetng we agreed that lazily loading source maps was the right approach to pursue.

PR #383, opened 2019-01-07
build: fix Windows builds

kinwa91:


nolanmar511:
@kinwa91 -- Could you explain what autorelease does?

With pre-compiled binaries being added to this module, I'm think we'll have to manage our own releases, but I'm curious to know more about what autorelease does.

kinwa91:
I see. So we are thinking about publishing locally then?

nolanmar511:
I think we're planning to publish using kokoro, still -- just our kokoro release will have to be different, since we'll want to create, test, and publish binaries before we release to npm in our job.

kinwa91:
I see - I'll hold off merging this until we have the Kokoro release ready.

kinwa91:
@nolanmar511 The Windows build is finally working, but is failing. Do you have any insights into what's causing it?

.kokoro/test.bat
```diff
@@ -12,16 +12,7 @@
...
 
-call npm install -g npm@latest || goto :error
-call npm install || goto :error
-call npm run test || goto :error
-
-goto :EOF
-
-:error
-exit /b 1
+"C:\Program Files\Git\bin\bash.exe" github/cloud-profiler-nodejs/.kokoro/test.sh
```
nolanmar511: I think this needs to point to github/cloud-profiler-nodejs/.kokoro/test.bat
(instead of .sh)?

.kokoro/publish.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
nolanmar511: I don't know that we want this file for cloud-profiler-nodejs yet.

.kokoro/test.bat
nolanmar511: I just made this change, because I'm curious to see if windows tests will pass.

nolanmar511:
Tests failing are related to source maps, so are probably failing because file paths look different on windows/linux. I'll take a closer look.

nolanmar511:
@kinwa91 -- I deleted publish.sh.

Any problems with the changes I've made?

kinwa91:
LGTM

kinwa91:
I also removed CircleCI, officially

nolanmar511:
Thanks!

PR #384, opened 2019-01-09
v0.3.0

nolanmar511:
Release notes: https://github.com/googleapis/cloud-profiler-nodejs/releases/edit/untagged-0029200f20cca0660243

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/profiler",
-  "version": "0.2.3",
+  "version": "0.3.0",
   "description": "Adds support for Stackdriver Profiler to node.js applications",
   "repository": "googleapis/cloud-profiler-nodejs",
   "main": "out/src/index.js",


ofrobots:
AFAICT, there are no breaking changes. Can you make that explicit in the release notes?

Secondly, is there a reason not to go to 1.0.0 (i.e. API stable) at this point. It seems that this module has a stable user-exposed API. Once you have 1.0.0, people can derive meaningful information just from the version for a given release. In the 0.x.y, people have to read the release notes carefully to understand whether they would expect any breakage.

aalexand:
@ofrobots @nolanmar511 I am sorry, I merged the PR by mistake. We should probably revert since there are comments from ofrobots.

nolanmar511:
NP -- I have a PR to revert this one.

Once that's in 

I think the next release can be 1.0.0.

The only thing that's been holding me back from 1.0.0, at least, has been a desire to get a bit more feature-complete first -- so have source map support finalize and have at least some pre-compiled binaries. But, it's probably not reasonable to wait for the agent to be more feature-complete before going to 1.0.0, especially given that I don't expect the API to change.

ofrobots:
In semver, (especially in the npm ecosystem), 1.0.0 is not about feature completeness, it is about setting expectations with your users that you are willing to use the version to communicate the semver impact of the changes. In fact, the recommended practice is to *start* with 1.0.0 for the first version.

PR #388, opened 2019-01-10
chore: update @google-cloud/common

nolanmar511:
Manually tested to confirm that profiling agent still backs off as expected when many agents are running.


JustinBeckwith:
@nolanmar511 just checking in - this change should have the fix to pass the full response back during errors.  Did we get that fixed up properly for y'all?

PR #400, opened 2019-02-08
feat: start publishing pre-compiled binaries

nolanmar511:
Part of #47.

After this change, we will release Node.js by bumping the version on github, then manually triggering a kokoro release build. Precompiled binaries for non-alpine linux will be generated.

I bump the version to an alpha version so, after this is merged, I can create and test a release.


JustinBeckwith:
Does this mean that y'all can use [releasetool](https://github.com/googleapis/releasetool/) and fit in with the process of all the other modules!?  @kinwa91  That would be awesome!

nolanmar511:
Failing tests are not related to changes in this PR.

PR #405, opened 2019-02-14
build: use linkinator for docs test

JustinBeckwith:


nolanmar511:
Merging.
"npmaudit" and "System test grpc-js" both failed prior to this change.

PR #407, opened 2019-02-14
chore: copy continuous and presubmit kokoro configs into separate directories

nolanmar511:
This is the first step to moving the presubmit and continuous configs into their own directories.


aalexand:
@nolanmar511 @JustinBeckwith More things failing?

nolanmar511:
There are some unexpected failures here -- they seem to be kokoro-related (test statuses suggets using kokoro:force-run). 
Re-running the test to see if that fixes things.

nolanmar511:
All tests except "System test grpc-js" are now passing.

PR #413, opened 2019-02-19
chore: finish separating kokoro configs into multiple directories

nolanmar511:
Finishes changes started in #407.

Also adds common configs to each directory with kokoro configs.

### DIFF
--- a/tools/kokoro/continuous/common.cfg
+++ b/tools/kokoro/continuous/common.cfg
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


aalexand:
@nolanmar511 There is Kokoro build failure?

nolanmar511:
The failure was real. Fixed by adding common.cfg to directories for presubmit, continuous, and release configs.

nolanmar511:
PTAL

nolanmar511:
I'm merging this change to reduce noise, since it should fix continuous test failures we're seeing.

PR #414, opened 2019-02-20
DO NOT LAND: direct pprof support

ofrobots:
Experimenting with a way to get this module to generate pprof output directly.

### DIFF
--- a/ts/src/cli.ts
+++ b/ts/src/cli.ts
@@ -1,0 +1,29 @@
+import { writeFileSync, fstat } from 'fs';
+import { gzipSync } from 'zlib';
+import {TimeProfiler} from './profilers/time-profiler';
+import {perftools} from '../../proto/profile';
+
+const DEFAULT_INTERVAL_MICROS = 1000;  // 1ms.
+
+let timeProfiler: TimeProfiler;
+function profile(intervalMicros = DEFAULT_INTERVAL_MICROS) {
+  if (timeProfiler) {
+    throw new Error('already profiling');
+  }
+  timeProfiler = new TimeProfiler(intervalMicros);
+  timeProfiler.startProfiling();
+}
+
+process.on('exit', code => {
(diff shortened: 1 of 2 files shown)


ofrobots:
So it would be good if we could split the `src/profilers` directory + native bindings into a separate module and expose the `start` and `stop` functions directly. The control, options and auth logic for profiling can continue to remain in the this module; which will become a pure-js module.

nolanmar511:
+1 to splitting `src/profilers` directory + native bindings into a separate module and expose the `start` and `stop` functions directly.

I would really love to get to the point where we could have have a module that, like go's ["net/http/pprof" package](https://golang.org/pkg/net/http/pprof/).

ofrobots:
Any suggestions for the name of the extracted pprof-based profling module?

- `pprof`: may be confused with the pprof tool itself?
- `pprof-profiler`: seemingly redundant name seems redundant. Hopeful the first `p` doesn't stand for `profiler` as well :)
- `node-pprof`

aalexand:
Splitting seems reasonable. For naming, how about `pprof-nodejs`? Seems to play well with `cloud-profiler-nodejs`.

ofrobots:
@aalexand for the repo name that would work. But for the npm module name, the `nodejs` part would be redundant. I personal vote is for `pprof` as the module name. The confusion can be managed in the README.

aalexand:
@ofrobots SGTM. Would the repo be googleapis/pprof-nodejs or google/pprof-nodejs? The latter is closer to google/pprof.

ofrobots:
IMO it should be `google/pprof-nodejs`. The module would add pprof profiling support for Node.js. It doesn't have anything to do with Google APIs, so it makes sense for it to be in the google github org.

aalexand:
I agree.

ofrobots:
Moved to https://github.com/google/pprof-nodejs/pull/8.

PR #416, opened 2019-02-20
fix: correct errors in kokoro release

nolanmar511:
In experimenting with the kokoro release, I found I that there are some errors in current kokoro configs and scripts used.

Right now, we want to call the publish script directly (rather than use devrel's trampoline.sh, which we aren't currently set up to do anyways). Also the release workflow did not copy binaries to "gs://cloud-profiler/nodejs/release".


aalexand:
@nolanmar511 The first paragraph in the PR description is incomplete.

PR #418, opened 2019-02-21
build: add grpc-js system-test to continuous

### LINKED ISSUE #403: system-tests grpc-js fail
This library does not use gax so should not have system tests grpc-js at all, but if we add that task here to make the setup similar, its configuration must be fixed. Right now the task is failing immediately with

```
[16:37:46] [ID: 5735578] Build finished after 1 secs, exit value: 0
[16:37:46] 
[16:37:46] 
[16:37:46] Command done in 10 seconds
[16:37:49][ERROR] Failed to get build config
com.google.devtools.kokoro.config.ConfigException: Couldn't find build configuration under /tmp/workspace/workspace/cloud-devrel/client-libraries/nodejs/continuous/googleapis/cloud-profiler-nodejs/node8/system-test-grpcjs/src/github/cloud-profiler-nodejs/.kokoro/continuous/node8.
```

@kinwa91 can you please take a look and fix? Thanks!

jkwlui:
Closes #403

### DIFF
--- a/.kokoro/continuous/node8/system-test-grpcjs.cfg
+++ b/.kokoro/continuous/node8/system-test-grpcjs.cfg
@@ -1,0 +1,12 @@
+# Download resources for system tests (service account key, etc.)
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/google-cloud-nodejs"
+
+env_vars: {
+    key: "TRAMPOLINE_BUILD_FILE"
+    value: "github/cloud-profiler-nodejs/.kokoro/system-test.sh"
+}
+
+env_vars: {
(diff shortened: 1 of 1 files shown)


aalexand:
@jkwlui Could you please remind why we want to have this test as part of cloud-profiler-nodejs? This repo does not use gRPC.

jkwlui:
@aalexand We are aware that cloud-profiler-nodejs doesn't use gRPC. But we've decided to keep the configs here so that it makes our CI config consistent across Node.js repositories in googleapis.

callmehiphop (approved):

Issue #315, opened 2018-10-22
Stop suppressing warnings with -Wno-cast-function-type


aalexand:
```
Mon Oct 22 15:22:36 UTC 2018: ++ node -e 'console.log(require('''./package.json''').version);'
Mon Oct 22 15:22:36 UTC 2018: + VERSION=0.2.2
Mon Oct 22 15:22:36 UTC 2018: + PROFILER=/cloud-profiler-nodejs/google-cloud-profiler-0.2.2.tgz
Mon Oct 22 15:22:36 UTC 2018: + TESTDIR=/test
Mon Oct 22 15:22:36 UTC 2018: + mkdir -p /test
Mon Oct 22 15:22:36 UTC 2018: + cp -r testing/busybench /test
Mon Oct 22 15:22:36 UTC 2018: + cd /test/busybench
Mon Oct 22 15:22:36 UTC 2018: + retry npm install --nodedir=/.nvm/versions/node/v10.12.0 /cloud-profiler-nodejs/google-cloud-profiler-0.2.2.tgz
Mon Oct 22 15:22:36 UTC 2018: + for i in {1..3}
Mon Oct 22 15:22:36 UTC 2018: + npm install --nodedir=/.nvm/versions/node/v10.12.0 /cloud-profiler-nodejs/google-cloud-profiler-0.2.2.tgz
Mon Oct 22 15:22:40 UTC 2018: In file included from ../../../nan/nan_new.h:189:0,
Mon Oct 22 15:22:40 UTC 2018: from ../../../nan/nan.h:222,
Mon Oct 22 15:22:40 UTC 2018: from ../bindings/sampling-heap-profiler.cc:19:
Mon Oct 22 15:22:40 UTC 2018: ../../../nan/nan_implementation_12_inl.h: In static member function ‘static Nan::imp::FactoryBase<v8::StringObject>::return_t Nan::imp::Factory<v8::StringObject>::New(v8::Local<v8::String>)’:
Mon Oct 22 15:22:40 UTC 2018: ../../../nan/nan_implementation_12_inl.h:340:37: warning: ‘static v8::Local<v8::Value> v8::StringObject::New(v8::Local<v8::String>)’ is deprecated: Use Isolate* version [-Wdeprecated-declarations]
Mon Oct 22 15:22:40 UTC 2018: return v8::StringObject::New(value).As<v8::StringObject>();
Mon Oct 22 15:22:40 UTC 2018: ^
[...]

PR #424, opened 2019-02-25
chore: silence deprecation warnings that appear with gcc 8

nolanmar511:
https://github.com/nodejs/nan/issues/807#issuecomment-455750192 recommends using the "-Wno-cast-function-type" to silence deprecations warnings that appear with GCC 8.

Silences warnings users see in #315.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -5,7 +5,8 @@
       "sources": [ 
         "bindings/profiler.cc",
       ],
-      "include_dirs": [ "<!(node -e \"require('nan')\")" ]
+      "include_dirs": [ "<!(node -e \"require('nan')\")" ],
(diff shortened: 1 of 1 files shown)


binding.gyp
```diff
@@ -5,7 +5,8 @@
       "sources": [ 
         "bindings/profiler.cc",
       ],
-      "include_dirs": [ "<!(node -e \"require('nan')\")" ]
+      "include_dirs": [ "<!(node -e \"require('nan')\")" ],
+      "cflags": [ "-Wno-cast-function-type" ]
```
aalexand: It would be good to add a comment here on why it's added.

nolanmar511: Thanks!
https://github.com/googleapis/cloud-profiler-nodejs/pull/428

PR #426, opened 2019-02-26
refactor: extract native code into externa module

ofrobots:
The native bindings have been moved to the pprof module. This module
can be pure-JS now.

Opening this early to get a code-review. This needs to wait for `pprof` to be published to npm before this can be merged.


ofrobots:
@nolanmar511 this is ready for review.

nolanmar511:
Kokoro continuous e2e test fails because build-essentials is not installed on the VM.

I will need to re-work continuous/presubmit integration tests now that we're pulling out pprof-nodejs.

nolanmar511:
LGTM pending the e2e tests passing. 

I'm updating the e2e tests so they should pass in #430.

PR #429, opened 2019-02-28
fix: fix testing bash script so code coverage reported

nolanmar511:
Currently, code coverage is not run. This is the error we see:
```
+ '[' false -ne true ']'
github/cloud-profiler-nodejs/.kokoro/test.sh: line 42: [: false: integer expression expected
```

### DIFF
--- a/.kokoro/test.sh
+++ b/.kokoro/test.sh
@@ -39,7 +39,7 @@ WINDOWS="false"
     ;;
   esac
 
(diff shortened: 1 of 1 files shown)


nolanmar511:
Code coverage was uploaded: https://codecov.io/gh/googleapis/cloud-profiler-nodejs/pull/429

PR #434, opened 2019-03-08
feat: support Knative environment variables for service and version

### LINKED ISSUE #308: Support Knative environment variable for service and version
The Knative open source project injects in the user container the variables `K_SERVICE` and `K_REVISION` that contain the equivalent of `GAE_SERVICE` and `GAE_VERSION`.

The profiler should pick them up if they are available and map them to service and version context.

nolanmar511:
Fixes #308

### DIFF
--- a/ts/test/test-init-config.ts
+++ b/ts/test/test-init-config.ts
@@ -369,7 +369,7 @@ describe('createProfiler', () => {
        assert.deepEqual(profiler.config, expConfig);
      });
 
-  it('should get values from  knative environment variable when not specified in config or other environment variables',
+  it('should get values from Knative environment variables when values not specified in config or other environment variables',
      async () => {
        process.env.K_SERVICE = 'k-service';
        process.env.K_REVISION = 'k-version';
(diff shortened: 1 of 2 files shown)


ts/test/test-init-config.ts
```diff
@@ -369,6 +369,45 @@ describe('createProfiler', () => {
        assert.deepEqual(profiler.config, expConfig);
      });
 
+  it('should get values from Knative environment variables when values not specified in config or other environment variables',
+     async () => {
+       process.env.K_SERVICE = 'k-service';
```
aalexand: It would be good to get confirmation from @steren that K_SERVICE won't ever contain uppercase letters and, more generally, will always match the profiler's requirements for service name strings.

steren: K_service follows the Kubernetes resource naming requirements. As far a I understood, uppercase is not allowed.

nolanmar511: It looks like kubernetes might allow starting a name K_SERVICE with a number ([this kubernetes doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) says "By convention, the names of Kubernetes resources should be up to maximum length of 253 characters and consist of lower case alphanumeric characters, -, and ., but certain resources have more specific restrictions."), but otherwise should be safe.

I think it's probably still a worthwhile change.

PR #436, opened 2019-03-19
fix: profiler should still start when source mapper can't be initialized

nolanmar511:
When running a benchmark in docker, Jianqiao saw this error:
```
@google-cloud/profiler Failed to initialize source maps and start profiler: Error: failed to get source maps from /: Error: ENOENT: no such file or directory, lstat 
```

Lazy source map loading will likely change this, but as a work-around for now, I don't think failure to initialize the SourceMapper should disabled the profiler.

### DIFF
--- a/ts/src/profiler.ts
+++ b/ts/src/profiler.ts
@@ -331,8 +331,9 @@ export class Profiler extends ServiceObject {
             await SourceMapper.create(this.config.sourceMapSearchPath);
       } catch (err) {
         this.logger.error(
-            `Failed to initialize source maps and start profiler: ${err}`);
-        return;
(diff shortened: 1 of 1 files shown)


ts/src/profiler.ts
```diff
@@ -331,8 +331,9 @@ export class Profiler extends ServiceObject {
             await SourceMapper.create(this.config.sourceMapSearchPath);
       } catch (err) {
         this.logger.error(
```
aalexand: Should this be a warning now?

kalyanac: I think error is the right level because a feature has been disabled automatically. It is partially catastrophic for the user. But both are fine, slight preference for error.

aalexand:
Sounds good.

PR #438, opened 2019-03-19
chore: use wombat to publish to npm

jkwlui:


nolanmar511:
Thanks so much!!!
If you could point me to how to use this to publish npm (once this change is ready), that would be awesome!

PR #442, opened 2019-03-27
chore: use version of pprof module with pre-compiled binaries

nolanmar511:


package.json
```diff
@@ -43,7 +43,7 @@
     "p-limit": "^2.0.0",
     "parse-duration": "^0.1.1",
     "pify": "^4.0.0",
-    "pprof": "^0.1.0",
+    "pprof": "^0.2.0-alpha",
```
aalexand: This won't affect customers of the module?

nolanmar511: Sorry, I hadn't pushed a commit; I'd originally tested that the pprof module was working as expected here before publishing a production version of the module. 

This change will impact customers (they'll start using pre-built binaries if they're on linux or alpine).

PR #445, opened 2019-03-28
v1.1.0

macieksmuga:
http://go/gh/googleapis/cloud-profiler-nodejs/releases/edit/untagged-cc97ad0b6ffef47c6450

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/profiler",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "description": "Adds support for Stackdriver Profiler to Node.js applications",
   "repository": "googleapis/cloud-profiler-nodejs",
   "main": "out/src/index.js",


nolanmar511:
Release process was completed, but agent release was blocked in CircleCI because of npm audit failure.

Asked about when mocha would have another release [here](https://github.com/mochajs/mocha/issues/3842). 

For now, we still have a kokoro release workflow, and I plan to trigger that to try to release the agent, since I'm not sure how to work around the CircleCI failure without sending another commit (and then our tags for commits would be a bit messier).

nolanmar511:
Kokoro release workflow failed to publish (that workflow was never well tested, so not surprising); I'll look at triggering through CircleCI.

nolanmar511:
I got this released using a separate branch, where I switched to using wobmat-dressing-room to publish, and the associated npm token. 

I think transitioning to using https://github.com/googleapis/releasetool (I'll need to add back a few kokoro configs, but should be fairly quick) is the best way forward.

PR #454, opened 2019-04-02
docs: mention building binaries and prebuilt binaries in docs

nolanmar511:


README.md
```diff
@@ -28,9 +28,23 @@ profiler is not stable with those versions of Node.js.
...
+component that is used to collect profiles with v8's CPU and Heap profilers.
+You may need to install additional dependencies to build the `pprof` module.
+    * For Linux: `pprof` has prebuilt binaries available for Linux and Alpine
+    Linux for Node 6, 8, 10, and 11. No additional dependencies are required.
+    * For other environments: when using `@google-cloud/profiler` on environments
+    that `pprof` does not have prebuilt binaries for, the module
+    [`node-gyp`](https://www.npmjs.com/package/node-gyp) will be used to
+    build binaries. See `node-gyp`'s
+    [documentation](https://github.com/nodejs/node-gyp#installation)
+    for information on dependencies required to build binaries with `node-gyp`.
```
kalyanac: It will be helpful to mention that Python2, build-essential are needed if prebuilt binaries are not available for their environment. There is mention of built-essential in a later section but doesnt say why

nolanmar511: Thanks -- I've removed references to build-essential below, now that pre-built binaries are available (so build-essential should not be needed). 

I don't know that I want to consistently track node-gyp's dependencies across platforms, so I linked to its documentation. Should I instead be more explicit and note the dependencies on Unix/MacOS/Windows?

kalyanac: Agreed, deferring to node-gyp documentation is better.

kalyanac:
ignoring failed npmaudit for now

PR #466, opened 2019-04-26
fix:  update dependencies and tsconfig to support Node 6

nolanmar511:


nolanmar511:
Clarified the title.
We have continuous integration tests which use Node 6, and those are failing. 
I want to get those tests passing, the go through a more conscious effort to drop support for Node 6.

nolanmar511:
Closing this.

It's probably just too hard to support Node 6, and Node 6 is basically end of life anyways.

nolanmar511:
Actually, want to get this change in. 
It will give time for a more graceful wrap-up of Node 6 support.

PR #468, opened 2019-04-26
fix: update dependencies for npm audit

nolanmar511:


kalyanac:
ignoring the node6 failures for now.

PR #473, opened 2019-04-29
update to .nycrc with --all enabled

bcoe:
to work properly, nyc should also be updated to 14


bcoe:
@nolanmar511 :+1: I'm about to send an automated PR for all repos, which might recreate this PR.

PR #475, opened 2019-05-02
build!: upgrade engines field to >=8.10.0

bcoe:
BREAKING CHANGE: drops Node 6 from engines


package.json
```diff
@@ -96,6 +96,6 @@
     ]
   },
   "engines": {
-    "node": ">=6.12.3 <8.0.0 || >=8.9.4 <10.0.0 || >=10.4.1"
+    "node": ">=8.10.0"
   }
-}
+}
```
nolanmar511: Nit: Could you add back the new line at the end of the file?

package.json
```diff
@@ -96,6 +96,6 @@
     ]
   },
   "engines": {
-    "node": ">=6.12.3 <8.0.0 || >=8.9.4 <10.0.0 || >=10.4.1"
+    "node": ">=8.10.0"
```
nolanmar511: For this agent, I think "engines" would need to be:
">=8.10.0 <10.0.0 || >=10.4.1"

Also, What's the reason for no longer supporting 8.9.X?

PR #476, opened 2019-05-02
build!: upgrade engines field to >=8.10.0

bcoe:
This drops support for all Node.js versions < 8.10.0, please upgrade

BREAKING CHANGE: library now requires Node >= 8.10.0


package.json
```diff
@@ -96,6 +96,6 @@
     ]
   },
   "engines": {
-    "node": ">=6.12.3 <8.0.0 || >=8.9.4 <10.0.0 || >=10.4.1"
+    "node": ">=8.10.0"
```
nolanmar511: For context, why is support for 8.9.X dropped?

Also, for this module, the engine would need to be ">=8.10.0 <10.0.0 || >=10.4.1". Some versions of Node 10 are not compatible with the profile.

PR #477, opened 2019-05-02
Revert "build!: upgrade engines field to >=8.10.0"

nolanmar511:
Reverts googleapis/cloud-profiler-nodejs#476

#476 should not have been merged in. 

It allows the profiler to run with versions of Node 10 impacted by https://bugs.chromium.org/p/chromium/issues/detail?id=847863 (gc can take several minutes when heap profiling is enabled)


nolanmar511:
No problems!

FYI to bcoe@ -- PR#476 has been reverted.

PR #491, opened 2019-05-21
chore: retry npm install in system test when the command hangs 

nolanmar511:


system-test/integration_test.go
```diff
@@ -93,7 +93,7 @@ retry git fetch origin {{if .PR}}pull/{{.PR}}/head{{else}}{{.Branch}}{{end}}:pul
 git checkout pull_branch
 git reset --hard {{.Commit}}
 
-retry npm install --nodedir="$NODEDIR" &>/dev/ttyS2
+retry node $HOME/cloud-profiler-nodejs/system-test/npm-install-retry.js 60000 3 install --nodedir="$NODEDIR" &>/dev/ttyS2
```
aalexand: Do you truly mean to retry the already retrying command?

nolanmar511: Good catch!
Switched to using the approach you'd suggested that uses timeout.

aalexand:
@nolanmar511 The PR title needs some polish. Code lgtm.

PR #492, opened 2019-05-22
DO NOT REVIEW: trying to run system test with different account

nolanmar511:


ofrobots:
For PRs that are not ready for review, might I suggest using branches rather than PRs? Once the CI and experimentation is green, they can be converted to PRs. This may cut down on email.

CircleCI will branches on the repo, but I am less certain about Kokoro. If it is a bother to get this working correctly, it is no biggie. It is probably not worth figuring out how to do this with kokoro if this doesn't work out of the box / trivially.

nolanmar511:
I created this PR to run the Kokoro presubmit tests (background is somewhat long, but I was trying to check if Kokoro VMs might have permission to access a pre-created GCP project, since the project seemed to grant some permissions to Kokoro-related service accounts).

In the past, I had run continuous tests against a branch, but continuous testing failures on non-master branches were not obviously distinguishable from continuous testing failures against master, and ends up being somewhat disruptive.

PR #493, opened 2019-05-23
refactor: drop dependency on pify

JustinBeckwith:
Now that we dropped support for node.js 6, we can just use `util.promisify`.


nolanmar511:
The `ci/kokoro: docs` failure is unrelated to this change.

@jkwlui -- any ideas? It looks like it might be a test that's set up internally with Kokoro, but the build config isn't yet in this repo, since it failed with:
```
 Failed to get build config
com.google.devtools.kokoro.config.ConfigException: Couldn't find build configuration under /tmp/workspace/workspace/cloud-devrel/client-libraries/nodejs/presubmit/googleapis/cloud-profiler-nodejs/node10/docs/src/github/cloud-profiler-nodejs/.kokoro/presubmit/node10
```

PR #494, opened 2019-05-23
fix(deps): remove unused dependencies

JustinBeckwith:
From what I could tell, none of this stuff was being used.  Please do let me know if we need to keep any of it!


nolanmar511:
`ci/kokoro: docs` failure is unrelated to this change.

PR #509, opened 2019-06-14
feat(breaking!): support API endpoint override

JustinBeckwith:


nolanmar511:
There are two things I'll need to be careful about with respect to this change:

1. Updating @google-cloud/common has broken our backoff when multiple agents with the same deployment are running. We don't have automated tests for this, and I'll need to test this manually.

2. It looks like this change is a breaking change for our agent (since setting baseApiUrl no longer has an effect; I know our probers that run against our test environment will be broken by this). Since we recently bumped out major version (for dropping Node 6), merging this definitely requires careful consideration for us. I don't want to need to bump our major version too frequently.

nolanmar511:
Added a "do not merge" label because I'd like to get a release of @google-cloud/profiler out before merging this change.

PR #516, opened 2019-06-27
feat(breaking!): require the project ID to be set before starting the profiling agent

nolanmar511:
Require the project ID to be set before starting the profiling agent. In some environments such as Istio, GCP metadata server may not be accessible for a few seconds after application start up. This will prevent the agent from starting when project ID is not provided and cannot be determined from the GCP metadata server.


kalyanac:
Any tests on an env that exercises the new code?

nolanmar511:
The E2E test on GCE requires that the project ID be fetched from metadata (but that wouldn't exercise the retries, generally).

kalyanac:
We talked offline about stopping the heap profiler if initialization fails. Will you update the code to do that?

aalexand:
I'd expect the PR description to clearly state the motivation for the change but it doesn't seem to? No links to fixed bugs nor explicit mentioning of "why" rather than "what".

aalexand:
"by at least 1.5 seconds." - where is this number coming from?

ts/src/config.ts
```diff
@@ -144,10 +144,17 @@ export interface Config extends GoogleAuthOptions {
   // All locations in profiles will reference locations in the running
   // JavaScript.
   disableSourceMaps?: boolean;
+
+  // Number of times to retry querying GCP metadata for the instance, zone,
+  // and projectId.
+  metadataRetries?: number;
+
+  // Milliseconds after failed query to GCP metadata to wait before retrying.
```
aalexand: "Delay after"

ts/src/index.ts
```diff
@@ -112,27 +108,56 @@ function initConfigLocal(config: Config): ProfilerConfig {
   return mergedConfig;
 }
 
+async function getGcpMetadata(
```
aalexand: "Gcp" or "GCP"? Not sure how the abbreviation are usually spelled in Node.js code, make sure it's consistent across.

ts/src/config.ts
```diff
@@ -144,10 +144,17 @@ export interface Config extends GoogleAuthOptions {
   // All locations in profiles will reference locations in the running
   // JavaScript.
   disableSourceMaps?: boolean;
+
+  // Number of times to retry querying GCP metadata for the instance, zone,
+  // and projectId.
+  metadataRetries?: number;
```
aalexand: Just to make sure we use "retry" terminology properly - these are retries, so the total maximum number of attempts to talk to the metadata server is 4, right? That is, 1 initial attempt + 3 retries.

ts/src/index.ts
```diff
@@ -112,27 +108,56 @@ function initConfigLocal(config: Config): ProfilerConfig {
...
+  try {
+    return await Promise.all([
+      gcpMetadata.project('project-id'),
+      gcpMetadata.instance('name'),
+      gcpMetadata.instance('zone'),
+    ]);
+  } catch (err) {
+    if (retriesLeft > 0) {
+      await delay(backoffMillis);
+      return getGcpMetadata(retriesLeft - 1, backoffMillis);
```
aalexand: Can we make it a loop rather than recursion?

Also, it seems that all three calls to the metadata server will be repeated if one of them fails? Doesn't seem right if so.

ts/test/test-init-config.ts
```diff
@@ -520,7 +562,89 @@ describe('createProfiler', () => {
       },
       disableSourceMapParams
     );
-    const profiler: Profiler = await createProfiler(config);
+    await createProfiler(config);
     assert.ok(!startStub.called, 'expected heap profiler to not be started');
   });
+  it('should retry specifed number of times to get project id from metadata', async () => {
```
aalexand: nit: id -> ID

ts/test/test-init-config.ts
```diff
@@ -520,7 +562,89 @@ describe('createProfiler', () => {
...
+    instanceMetadataStub = sinon.stub(gcpMetadata, 'instance');
+    instanceMetadataStub
+      .withArgs('name')
+      .resolves('gce-instance')
+      .withArgs('zone')
+      .resolves('projects/123456789012/zones/gce-zone');
+    projectMetadataStub = sinon.stub(gcpMetadata, 'project');
+    projectMetadataStub
+      .onCall(0)
+      .rejects('cannot get project id')
```
aalexand: id -> ID everywhere

ts/src/index.ts
nolanmar511: In Node.js, only the first letter of acronyms are capitalized when in 
 the middle of camel case variable names. 
So, this is consistent with variables names in the agent, and in libraries it depends on.

ofrobots: `getGcpMetadata` is correct. The js style guide would apply: go/js-style#naming-camel-case-defined (internal link).

ts/src/index.ts
```diff
@@ -112,27 +108,56 @@ function initConfigLocal(config: Config): ProfilerConfig {
   return mergedConfig;
 }
 
+async function getGcpMetadata(
+  retriesLeft: number,
+  backoffMillis: number
+): Promise<[string | undefined, string | undefined, string | undefined]> {
+  try {
+    return await Promise.all([
+      gcpMetadata.project('project-id'),
```
ofrobots: I believe `gcpMetadata.project` does retries underneath the covers already. There is some subtlety/complexity here between retrying on network vs. http errors, and I understand that gcpMetadata already has that logic.

nolanmar511: On certain environments, the metadata server is not available for some time (order of minutes) after start up.

Perhaps it makes sense to default to 0 retries, but I do think we need retries that are more configurable than gcp-metadata'a retries.

ts/src/config.ts
nolanmar511: Yep!
I added two test cases which (together) validate that the term "retries" is used correctly here.

ts/src/index.ts
kalyanac: ofrobots@ This change was motivated by the network delay in envoy startup. Envoy may not finish its initialization until a few seconds after the App has already started. Will ```gcpMetadata.project``` eventually fully support retrying for handling such environments? Since other agents will also face the same issue, will ```gcpMetadata.project```  support configurable retries? It will be cleaner and simpler to have it handled in one place than each agent having its own approach for retries.

ts/test/test-init-config.ts
nolanmar511: Done.

ts/test/test-init-config.ts
nolanmar511: Done.

ts/src/config.ts

ts/src/index.ts

ts/src/index.ts
ofrobots: Adding @JustinBeckwith . If there is a problem interacting with the metadata service, it will affect all the other client libraries as well. My preference would be to fix this in a central place. /cc @JustinBeckwith

nolanmar511:
@kalyanac  and @JustinBeckwith -- How do we want to handle network delays on Envoy start-up (kalyanac explained the motivation for this PR [in this comment](https://github.com/googleapis/cloud-profiler-nodejs/pull/516#discussion_r298742009) above)

ts/src/index.ts
kalyanac: @JustinBeckwith is there a plan for adding retries in the common libraries? And is there any WIP for supporting Istio?

JustinBeckwith:
@kalyanac I think this is a complex enough topic that we probably need to sync up on GVC with @bcoe.  The `gcp-metadata` module does this check with fail-fast and no-retries to ensure that code *not* running on GCP doesn't hang while determining the execution environment.  Starting a process that won't have network access, starting an auth flow, and then suddenly having network access is going to cause problems far beyond just this library, and feels like a pattern you *really* want to avoid.  I kinda reject the premise 🤣

kalyanac:
I completely agree about not blocking the app from starting. 
Have you considered how you will approach supporting Istio? We have done extensive testing with Istio and per our conclusions, there is no way around this than retrying. We provide user with the option to set retries so they can disable retries when not running on GCP/Anthos.

bcoe:
@kalyanac @JustinBeckwith there's a larger conversation already taking place around, in the case of gRPC libraries, moving retry logic out of clients and towards the API layer -- moving retry logic into `gcp-metadata` feels like it might step on the toes of this initiative.

Rather than retrying until metadata is available, could we fail by stubbing out a metrics client that then periodically tries to create an actual client? or do we want the process bootstrapping to block until envoy starts up?

A pattern I've seen in the past that can work well, is everything starts normally, and you'll then see a connection made to a service like envoy a few moments afterwards.

kalyanac:
In other agents, the implementation blocks till it has all the info it needs to initialize successfully. Node.js agent did not require all data to be available but that creates inconsistency in what data is reported in the profiles within the same profiling session, so we prefer to block till all necessary data is available. 

nolanmar@ is that accurate?

bcoe:
@kalyanac gotcha, I better understand now :+1: to summarize:

1. when containers are booted with Envoy proxy, there's a delay in `gcp-metadata` being able to access the network.
2. to circumvent this we've implemented retry logic in `cloud-profiler-nodejs`, _but_ this same issue would happen for any GCP API.

Seems like it might be a good call to implement this once in a library like `gcp-metadata` :+1: 

Out of curiosity, what advice do we give users to deal with this issue in their applications, i.e., if they're not running any GCP APIs, and this environment boots, will they fail to make connections to external APIs and services?

kalyanac:
Yes, on both counts. Other language agents for Profiler also implement retries. 

I am not sure I follow your last question fully. Users when running under Istio need to add necessary API endpoints to an allow list, and wait till Istio boots up to accept app traffic.

nolanmar511:
@kalyanac  -- I switch this change so it would just return an error, and the README with information on how to retry.

README.md
```diff
@@ -211,6 +252,8 @@ the service when starting the profiler:
         service: 'your-service',
         version: '1.0.0'
       }
+    }).catch((err) => {
+      console.log(`Failed to start profiler:  keyFilename: '/path/to/keyfile'
```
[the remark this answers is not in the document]
nolanmar511: Thanks! Done.

nolanmar511:
PTAL

README.md
```diff
@@ -58,7 +58,9 @@ to your [`package.json`](https://docs.npmjs.com/files/package.json#dependencies)
 2. Include and start the profiler at the beginning of your application:
 
     ```js
-    var profiler = require('@google-cloud/profiler').start();
+    var profiler = require('@google-cloud/profiler').start().catch((err) => {
```
ofrobots: This problem existed before your change, but it gets more confusing after you change: AFAICT, `start` returns `Promise<void>` so assigning it to a variable named `profiler` doesn't make too much sense to me. I would suggest removing the `var profiler = ` from this statement.

ts/src/index.ts
```diff
@@ -195,8 +218,7 @@ export async function start(config: Config = {}): Promise<void> {
   try {
     profiler = await createProfiler(config);
   } catch (e) {
-    logError(`${e}`, config);
-    return;
+    throw e;
```
ofrobots: Why catch at all? Omit the try/catch block and it will bubble to the caller.

README.md
```diff
@@ -86,7 +88,9 @@ configuration options. These options can be passed to the agent through the
 object argument to the start command shown below:
 
 ```js
-require('@google-cloud/profiler').start({disableTime: true});
+require('@google-cloud/profiler').start({disableTime: true}).catch((err) => {
```
ofrobots: Optional suggestion: instead of making all the examples verbose with a catch block, you could use something like this:

```js
const success = await require('@google-cloud/profiler').start({...});
```

The await succinctly implies that the returned value is a promise. A rejection on the awaited promise will correctly be turned into throw.

ofrobots: Maybe even omit the `const success = `.

README.md
nolanmar511: Thanks! Done.

README.md
nolanmar511: Thanks! Done.

nolanmar511:
PTAL

ts/src/index.ts
```diff
@@ -192,33 +215,17 @@ export async function createProfiler(config: Config = {}): Promise<Profiler> {
  */
 export async function start(config: Config = {}): Promise<void> {
   let profiler: Profiler;
-  try {
-    profiler = await createProfiler(config);
-  } catch (e) {
-    logError(`${e}`, config);
-    return;
-  }
+  profiler = await createProfiler(config);
```
ofrobots: No need to keep the declaration and assignment separate anymore. `const profiler = await createProfiler(config)`.

ts/src/index.ts
```diff
@@ -192,33 +215,17 @@ export async function createProfiler(config: Config = {}): Promise<Profiler> {
...
  */
 export async function startLocal(config: Config = {}): Promise<void> {
   let profiler: Profiler;
-  try {
-    profiler = await createProfiler(config);
-  } catch (e) {
-    logError(`${e}`, config);
-    return;
-  }
+  profiler = await createProfiler(config);
```
ofrobots: likewise.

ts/src/config.ts
```diff
@@ -171,6 +171,11 @@ export interface ProfilerConfig extends GoogleAuthOptions {
   disableSourceMaps: boolean;
 }
 
+// Interface for an initialized profiler config.
+export interface ProfilerConfig extends LocalConfig {
+  projectId: string;
```
ofrobots: Apologies if this was discussed earlier, and I missed it. Why are we requiring all users, even those not running on Istio to now start providing a `projectId`, when their code works fine as-is? With this breaking change, their code may stop compiling. In some cases they would need to go though significant hoops to provide a project id, wherein they didn't need to before.

Looking at the implementation, we actually are able to fetch the `projectId` from the metadata service on the users behalf, and use it even if the `projectId` was not specified in the config. IOW, this breaking change seems more strict than what the code actually requires.

Am I overlooking something?

kalyanac: We always required project ID. It should be either discoverable from metadata server, provided in a config file or provided as a init config. Previously, the code attempted to read project ID from metadata server on every profile upload. This causes a problem with Istio env. This PR should make the agent init wait till it can obtain the project ID from metadata server or fail. This should not break any existing code but will require a small change if users are already running on Istio.

nolanmar511: As additional reassurance that this is working as intended, our Kokoro system test uses a typescript benchmark that doesn't specify the project ID when starting profiler, and it continues to build and pass.

ts/src/index.ts
nolanmar511: Thanks! Done.

ts/src/index.ts
nolanmar511: Thanks! Done.

ts/src/config.ts
ofrobots: Thanks for clarifying that `ProfiledConfig` is not publicly exported to the users, but rather the user-visible config is `Config`.

PR #529, opened 2019-08-14
fix(dep): update to be compatible with @google-cloud/common 2.1.X

nolanmar511:


nolanmar511:
@JustinBeckwith -- Any guidance on making our headers compliant with header-check?

JustinBeckwith:
@nolanmar511 right now it's safe to ignore that - it shouldn't be required.  Working with @bcoe to figure out how to roll this bot out :)

PR #539, opened 2019-09-10
chore(deps): pin pprof 

nolanmar511:
`@google-cloud/profiler` depends heavily on the `pprof` module. 

We want to be sure that changes to `pprof` that introduce bugs or problems in `@google-cloud/profiler` can be caught by `@google-cloud/profiler`'s  presubmit tests, rather than immediately impacting `@google-cloud/profiler` as its running in prod. 

This will allow us to concentrate testing in this repository, rather than duplicate all testing here and in https://github.com/google/pprof-nodejs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "extend": "^3.0.2",
     "gcp-metadata": "^2.0.0",
     "parse-duration": "^0.1.1",
-    "pprof": "^1.0.0",
+    "pprof": "1.1.0",
     "pretty-ms": "^5.0.0",
     "protobufjs": "~6.8.6",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "extend": "^3.0.2",
     "gcp-metadata": "^2.0.0",
     "parse-duration": "^0.1.1",
-    "pprof": "^1.0.0",
+    "pprof": "1.1.0",
```
aalexand: @ofrobots for Ali's opinion.

I think even with pinning it's common to still allow revision version changes so that security patches can still be picked up.

PR #572, opened 2019-12-24
build: use c8 for coverage

JustinBeckwith:


bcoe:
@nolanmar511 is this safe to land with the circleci issues?

package.json
```diff
@@ -52,6 +51,7 @@
     "@types/pretty-ms": "^4.0.0",
     "@types/sinon": "^7.0.12",
     "@types/tmp": "0.1.0",
+    "c8": "^7.0.0",
     "codecov": "^3.0.0",
```
nolanmar511: I assume codecov is no longer needed. Could this and codecov.yaml be removed?

bcoe: @nolanmar511 codecov is the bin used to upload coverage results to the upstream service we use, so this is still needed (nyc and c8 are what collect the coverage locally).

nolanmar511: Okay! Thanks for the clarification!

nolanmar511:
I removed the CircleCI tests since they've been redundant since the Kokoro tests were added. 

All checks should pass now, and this should be safe to merge.

bcoe:
@nolanmar511 thank you :clap:

PR #578, opened 2020-01-13
chore: use go modules in integration test

nolanmar511:
Our E2E test is broken, because it depends on google.golang.org/api/compute/v1,  which depends on google.golang.org/api/option, which is currently broken for all users who are not using go modules (https://github.com/googleapis/google-api-go-client/issues/441).

This should allow our E2E test to pass again.

### DIFF
--- a/system-test/integration_test.sh
+++ b/system-test/integration_test.sh
@@ -23,14 +23,14 @@ export GCLOUD_TESTS_NODEJS_PROJECT_ID="cloud-profiler-e2e"
 export GCLOUD_TESTS_NODEJS_ZONE="us-east4-b"
 export GOOGLE_APPLICATION_CREDENTIALS="${SERVICE_KEY}"
 
-# Move test to go path.
-export GOPATH="$HOME/go"
-mkdir -p "$GOPATH/src"
-cp -R "system-test" "$GOPATH/src/proftest"
-
 # Run test.
-cd "$GOPATH/src/proftest"
-retry go get -t -d -tags=integration .
+cd "system-test"
+
+# Initializing go modules allows our dependencies to install versions of their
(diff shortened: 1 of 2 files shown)


system-test/integration_test.sh
```diff
@@ -23,13 +23,9 @@ export GCLOUD_TESTS_NODEJS_PROJECT_ID="cloud-profiler-e2e"
 export GCLOUD_TESTS_NODEJS_ZONE="us-east4-b"
 export GOOGLE_APPLICATION_CREDENTIALS="${SERVICE_KEY}"
 
-# Move test to go path.
-export GOPATH="$HOME/go"
```
aalexand: GOPATH is not needed when using modules? Maybe we should have stopped using it in the probers too then?

system-test/integration_test.sh
```diff
@@ -23,13 +23,9 @@ export GCLOUD_TESTS_NODEJS_PROJECT_ID="cloud-profiler-e2e"
...
-mkdir -p "$GOPATH/src"
-cp -R "system-test" "$GOPATH/src/proftest"
-
 # Run test.
-cd "$GOPATH/src/proftest"
+cd "system-test"
+go mod init
```
aalexand: Maybe add a comment to clarify what this achieves.

system-test/integration_test.sh
nolanmar511: Done.

system-test/integration_test.sh
nolanmar511: Done.

nolanmar511:
I'm merging this in. I'll deal with ci/kokoro: docs later; that failure was unrelated to this change.

PR #606, opened 2020-03-06
build: update linkinator config

JustinBeckwith:


JustinBeckwith:
@bcoe should we delete the `windows` tests we have running on kokorkokoroko?

PR #625, opened 2020-04-13
chore: clean-up tests and readme after dropping support for node 8

nolanmar511:
This change also removes integration tests for Node 11, and adds integration tests for Node 13.


bcoe:
@nolanmar511, you might consider this blurb I've been adding to most of the other libraries:

https://github.com/googleapis/google-cloud-node#supported-nodejs-versions

nolanmar511:
@bcoe -- Thanks for the suggestion!

I think I'll make the change in a follow-up -- what's the method for adding legacy tags?

PR #626, opened 2020-04-15
chore: mark cloud-profiler-nodejs as GA

nolanmar511:
This change also updates README.md to include a "Supported Node.js Versions" section.

The "Supported Node.js Versions" section and "Versioning" section were taken from [@google-cloud/common](https://github.com/googleapis/nodejs-common).

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,38 @@ the service when starting the profiler:
          }
        });
        ```
+
+## Supported Node.js Versions
+
+Our client libraries follow the [Node.js release schedule](https://nodejs.org/en/about/releases/).
+Libraries are compatible with all current _active_ and _maintenance_ versions of
+Node.js.
+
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
So this is great, but I wanted to present another option.  If you'd like to get this content, as well as any other changes we make broadly, you *could* un-ignore `README` from `synth.py`.  To keep custom content you want for the repo, you could put it in a partials file like this:
https://github.com/googleapis/nodejs-storage/blob/master/.readme-partials.yaml

I am happy to give this a shot if you like :)

kalyanac:
@JustinBeckwith 
Your suggestion SGTM. 
Let's wait to hear from @nolanmar511

JustinBeckwith:
I think this turned out really well actually:
https://github.com/googleapis/cloud-profiler-nodejs/pull/627

PR #632, opened 2020-04-16
chore: retry install go modules in integration test

nolanmar511:


system-test/integration_test.sh
```diff
@@ -29,7 +29,9 @@ cd "system-test"
 # Initializing go modules allows our dependencies to install versions of their
 # dependencies specified by their go.mod files. This reduces the likelihood of
 # dependencies breaking this test.
+go version
 retry go mod init e2e
```
aalexand, line 33: I don't think "go mod init" even needs to be retried - I believe it's a local command.

nolanmar511: Thanks!
Done.

PR #645, opened 2020-05-07
chore: compile then run golang test in e2e test

nolanmar511:
This allows the installation of dependencies for golang test to be retried.

### DIFF
--- a/system-test/integration_test.sh
+++ b/system-test/integration_test.sh
@@ -31,11 +31,11 @@ cd "system-test"
 # dependencies breaking this test.
 go version
 go mod init e2e
-retry go get -t -d -tags=integration .
+retry go test -c -tags=integration .
 
 if [ "$KOKORO_GITHUB_PULL_REQUEST_NUMBER" = "" ]; then
-  go test -timeout=30m -tags=integration -run TestAgentIntegration -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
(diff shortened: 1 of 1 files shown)


system-test/integration_test.sh
```diff
@@ -31,11 +31,11 @@ cd "system-test"
 # dependencies breaking this test.
 go version
 go mod init e2e
-retry go get -t -d -tags=integration .
+retry go test -c -tags=integration .
 
 if [ "$KOKORO_GITHUB_PULL_REQUEST_NUMBER" = "" ]; then
-  go test -timeout=30m -tags=integration -run TestAgentIntegration -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
+  ./e2e.test -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
```
kalyanac, line 37: Why was timeout dropped?

nolanmar511: The timeout is the timeout for running whatever is run by `go -test` (so, the test would timeout if compilation were slow, not if the test ran slowly); the timeout flag is not accepted by the next command. 

I'm not too concerned about dropping the timeout specified here. There is a timeout defined within the test ([here](https://github.com/googleapis/cloud-profiler-nodejs/blob/f69c7a73b17c150c2b523412e430b5d1ac03e12a/system-test/integration_test.go#L248)). I also think we specified a timeout here primarily to override the default timeout for `go test` (it's 10 minutes, documented [here](https://golang.org/pkg/cmd/go/internal/test/#pkg-variables)).

PR #651, opened 2020-05-28
chore: add backoff integration test

nolanmar511:


system-test/integration_test.sh
```diff
@@ -32,11 +32,13 @@ cd "system-test"
...
 go version
 go mod init e2e
-retry go test -c -tags=integration -timeout=30m -run=TestAgentIntegration .
+retry go get cloud.google.com/go/profiler/proftest@HEAD
+retry go test -c -tags=integration .
 
 if [ "$KOKORO_GITHUB_PULL_REQUEST_NUMBER" = "" ]; then
-  ./e2e.test -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO"
+   #TODO: -run_backoff_test=true is only for testing purposes; backoff tests should not run as part of presubmit tests.
+  ./e2e.test -commit="$COMMIT" -branch="$BRANCH" -repo="$REPO" -run_backoff_test=true
```
nolanmar511, line 40: I am having the backoff test run as a presubmit while the test is likely to change.

I must remove this before this change should be submitted.

kalyanac:
I see you assigned reviewers and removed the "do not review" from PR title. But the description and label seem like its not ready for review. Should I review this or wait?

nolanmar511:
I updated the description to clarify.

I want to run the presubmit test throughout the review process, and only modify [these lines](https://github.com/googleapis/cloud-profiler-nodejs/blob/8951c4011bb848792b10d3b84f92971f30a652b1/system-test/integration_test.sh#L39-L40) to stop that right before the PR is submitted. I'm trying to ensure we don't submit a broken continuous test.

sofisl (approved):

system-test/integration_test.go
```diff
@@ -34,20 +34,34 @@ import (
...
-	pr     = flag.Int("pr", 0, "git pull request to test")
+	repo           = flag.String("repo", "https://github.com/googleapis/cloud-profiler-nodejs.git", "git repo to test")
+	branch         = flag.String("branch", "", "git branch to test")
+	commit         = flag.String("commit", "", "git commit to test")
+	pr             = flag.Int("pr", 0, "git pull request to test")
+	runBackoffTest = flag.Bool("run_backoff_test", false, "Enables the backoff integration test. This integration test requires over 45 mins to run, so it is not run by default.")
 
 	runID             = strings.Replace(time.Now().Format("2006-01-02-15-04-05.000000-0700"), ".", "-", -1)
-	benchFinishString = "busybench finished profiling"
+	benchFinishString = "benchmark application(s) complete"
```
jqll, line 44: Just curious, is this consistent with the bench finish string for other agents? It might be good to be consistent, for example, it'll be easier to search logs.

jqll (approved):

system-test/busybench/src/busybench.ts
```diff
@@ -35,12 +35,14 @@ function busyLoop(durationSeconds: number) {
 }
 
 function benchmark(durationSeconds: number) {
-  // Allocate 16 MiB in 64 KiB chunks.
-  for (let i = 0; i < 16 * 16; i++) {
-    testArr[i] = new Array<number>(64 * 1024);
+  // Allocate arrMiB MiB in 1 MiB chunks.
+  for (let i = 0; i < arrMiB; i++) {
+    // (64 bits / number) * (1 byte / 8 bits) * 128 * 1024 = 1 MiB
```
kalyanac, line 40: 8 bytes * 127 * 1024 = 1 MiB

system-test/integration_test.go
nolanmar511: Yes (or, this will be true) --I'm switching all agents to this string as I add backoff tests.

system-test/busybench/src/busybench.ts
nolanmar511: Done.

nolanmar511:
The backoff test is no longer run as a presubmit test, so this change can be submitted.

kalyanac (approved):

Issue #697, opened 2020-10-07
Cloud Profiler has memory leak when used with v12.16.0 or greater and v14.0.0 or greater.


plegner:
Using the Cloud Profiler on AppEngine NodeJS Managed VMs causes a memory leak.

- Node.js version: `12.x`
- `@google-cloud/profiler` version: `^4.0.3`
- Environment: Appengine Managed VMs

Here is a memory profile with the cloud profiler enabled:

<img width="1393" alt="before-memory" src="https://user-images.githubusercontent.com/2439570/95300046-ebce4f00-087e-11eb-8185-f315d431eb64.png">

And here is the same server running with the cloud profiler disabled:

<img width="1387" alt="Screenshot 2020-10-07 at 09 22 10" src="https://user-images.githubusercontent.com/2439570/95300122-086a8700-087f-11eb-82d9-85c2fa85208a.png">

The only difference is commenting out these lines:

```
if (env === 'production') require('@google-cloud/profiler').start()
```

kalyanac:
Thank you for reporting this. Does this happen on one specific service or does it happen on other services as well? 

Do you run more than one instance of this service?

plegner:
I only have a single service where I've used the profiler, but it happens on all (2) instances running that service. Happy to send you the project/service ID.

(I had previously filed case #25366321 in Cloud support, and one of your colleagues suggested removing the profiler.)

nolanmar511:
I was able to reproduce running on my workstation, with Node.js 12.18.0 on my linux workstation using the AcmeAir application. 

Logging process.memoryUsage() shows a memory increase is in RSS, with no corresponding increases in heap total, heap used, external or array buffer categories. 

My next steps are to:
* reproduce using pprof-nodejs and a simpler-to-run application (this would allow us to confirm that the issue is with profile collection; it also generally allows us adjust profile collection settings to make the memory leak happen faster).
* identify if this is specific to a version of `@google-cloud/profiler` or more specific to a version of Node.js.
* identify if this leak is specific to the time or heap profiles.

nolanmar511:
@plegner -- Do you happen to know the specific version of Node.js being used?

I'm running an application with node 12.0.0 for comparison; so far, it looks like there is a memory leak when I use node v12.18.0, but not when I use node v12.0.0.

nolanmar511:
There does appear to be a chromium issue for this.
[...]

PR #699, opened 2020-10-12
fix: Update engines to prevent agent from geing used when it will cause memory leak.

nolanmar511:
Related to https://github.com/googleapis/cloud-profiler-nodejs/issues/697

Node versions 12.16.0 and greater are impacted by a memory leak in V8 profilers. 
https://bugs.chromium.org/p/v8/issues/detail?id=10883


kalyanac:
Please include a link to the V8 issue in the PR description. Also, file an issue to reenable versions >12.16 after the V8 issue is fixed.

bcoe (approved):

PR #706, opened 2020-10-30
Revert "fix: Update engines to prevent agent from getting used when it will cause memory leak."

nolanmar511:
Reverts googleapis/cloud-profiler-nodejs#699

A version of pprof-nodejs which works around the memory leak is being released, so this change is no longer needed.


macieksmuga (approved):

system-test/integration_test.go
```diff
@@ -289,7 +289,7 @@ func TestAgentIntegration(t *testing.T) {
 			},
 			name:          fmt.Sprintf("profiler-test-node13-%s-gce", runID),
```
kalyanac, line 290: name says node 13 but  nodeVersion is set to 12

kalyanac (approved):
Please Address the comment before merging.

system-test/integration_test.go
nolanmar511: Done.

nolanmar511:
I deleted Node 13 test.
We don't have prebuilt binaries for that version, so that test failed.

PR #709, opened 2020-11-02
feat: add support for Node 14

nolanmar511:


.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 13]
+        node: [10, 12, 13, 14]
```
JustinBeckwith, line 14: As part of this, maybe we could remove 13, and add 15?  Or separate PRs?

nolanmar511: I'll probably do this in separate PR. 

I'm mainly adding this PR to reflect that the pprof dependency now has node 14 pre-built binaries (and so, I want our next release to be a minor version bump and the Node 14 binaries to be noted in its release notes); but our transition in updating tests for pprof-nodejs with the Node 14 release was interrupted by work to fix the memory leak, so pprof-nodejs's testing is quite to the point where we should add node 14 support here.

bcoe (approved):

PR #740, opened 2021-06-04
test: display environment variables when encountering an error

florinpapa:


florinpapa:
Closing over concerns that the env variables could contain sensitive information.

PR #748, opened 2021-06-23
chore: clean npm cache on package installations

aalexand:
This should avoid persistent errors in case of rare npm cache corruptions.
Something like https://github.com/immerjs/immer/issues/546.


nolanmar511 (approved):

aalexand:
@nolanmar511 I just noticed this - thanks for fixing the PR title. I always forget using this semantic versioning compatible style.

PR #766, opened 2021-09-14
build: remove "Kokoro CI" required check

chingor13:
There is no such Kokoro job


nolanmar511:
The "Kokoro CI" test contains runs the Kokoro integration test under system-test/integration_test.go.

It seems like it was disabled in the rename of the main branch. This test is important to my team; I will work to re-enable it.

JustinBeckwith:
apologies @nolanmar511!  Thank you for fixing our mess here :/

PR #780, opened 2021-12-10
fix: log to console using structured logging

minherz:
Replace package `console-log-level` with `@google-cloud/logging-min`. Use LogSync class of the logging package to print structured logging to `stdout` to capture the correct severity level.
Also handle a case of printing multi-line log that is captured as multiple log entries.


minherz:
@bcoe can you help me to troubleshoot the reasons for check failures, please?

minherz:
Test fails will be fixed with https://github.com/mapbox/node-pre-gyp/issues/623

src/logger.ts
```diff
@@ -12,36 +12,69 @@
...
+  }
+
+  error(msg: string) {
+    if (this.severityThreshold > 0) {
+      this.log.error(this.log.entry(this.toOneLine(msg)));
+    }
+  }
+
+  private toOneLine(msg: string): string {
+    return msg.replace('\n', '\\n');
```
losalex, line 74: Do you need to handle '\r' as well?

src/logger.ts
```diff
@@ -12,36 +12,69 @@
...
+    }
+  }
+
+  error(msg: string) {
+    if (this.severityThreshold > 0) {
+      this.log.error(this.log.entry(this.toOneLine(msg)));
+    }
+  }
+
+  private toOneLine(msg: string): string {
```
losalex, line 73: Is one line a requirement?

src/logger.ts
```diff
@@ -12,36 +12,69 @@
...
-    level = 4;
+export class Logger {
+  private log: LogSync;
+  private severityThreshold: number;
+
+  constructor(readonly level?: number) {
+    if (level === undefined) {
+      level = defaultConfig.logLevel;
+    }
+    if (level < 0) {
```
losalex, line 40: I guess you should define a constants for this

losalex, line 40: Is this intentionally? I wonder in case of "invalid" value should you use `defaultConfig.logLevel`?

losalex (approved):
Some comments inside

src/logger.ts
minherz: I do not think that standalone `\r` should be handled. Would the following handle Windows new lines IYO:
```java
var temp = msg.replace('\r\n', '\\r\\n');
return temp.replace('\n', '\\n');
```

src/logger.ts
minherz: Yes. Printing to STDOUT is captured one line per log entry.

src/logger.ts
minherz: Can you be more specific? What values should be set to constants?

minherz: I am not sure what you mean here. If the level is undefined then the default level is used. The conditions for level value beyond the range are handled exactly the same way they were handled before the change. The change does not modify existing interfaces or behavior, it only replaces the engine that prints logs to STDOUT behind it.

losalex (approved):
Left some comments, see if you agree with those

src/logger.ts
minherz: It would change the current behavior.

src/logger.ts
minherz: I saw it a common practice of capturing multiline payload in Json is to replace with an anchor that can be used to pretty print the text later.
It is hard to restore the original multi-line string later (e.g. callstacks). Such work will require from a reader a lot of effort and manual work. With the anchor, a simple find-and-replace in an editor or with a script/code will do the work.
Would removing \r means that in certain OS (e.g. Windows) restoring will not work?

minherz: Just updated the code. Please, see the new version that handles new lines in both Window and Linux formats.

src/logger.ts
minherz: Added constants and a comment that explains where it comes from.

bcoe (approved):

PR #807, opened 2022-03-23
test: reduce integration test flakiness

amchiclet:
The test now retries when connecting to the metadata server.

Go 1.17 [build constraint](https://pkg.go.dev/go/build/constraint#pkg-overview) syntax was also added.

Tested by temporarily having presubmit also run the backoff tests.

The Golint check may fail because there is no .mod file. This is intentional and integration_test.sh does the module initialization.

Windows check may also fail but can be ignored ([Windows is not supported](https://cloud.google.com/profiler/docs/about-profiler#environment_and_languages)).

### DIFF
--- a/system-test/integration_test.go
+++ b/system-test/integration_test.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build integration && go1.7
 // +build integration,go1.7
 
 package e2e
@@ -96,7 +97,7 @@ git reset --hard {{.Commit}}
 
 retry npm_install --nodedir="$NODEDIR"
 
-npm run compile 
+npm run compile
 npm pack --nodedir="$NODEDIR" >/dev/null
 VERSION=$(node -e "console.log(require('./package.json').version);")
 PROFILER="$HOME/cloud-profiler-nodejs/google-cloud-profiler-$VERSION.tgz"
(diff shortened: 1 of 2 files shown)


nolanmar511 (approved):

system-test/integration_test.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build integration && go1.7
```
aalexand, line 15: Curious - what is the difference between these two lines?

```
//go:build integration && go1.7
// +build integration,go1.7
```

They look similar - to the point where I'm thinking "hmm, maybe one of them is redundant?"...

amchiclet: I think the minimal needed is `// +build` and things will still be working.

Go 1.16 will fail to build if only `//go:build` is used.

Go 1.17 and onward, it is recommended to use `//go:build`.

`gofmt` automatically adds `//go:build`, which is a safe thing to do.

But now that you've pointed it out, in integration_test.sh, the test is always built with Go 1.17, so having only the newer version would suffice.

Let me try that out and if the tests pass, I'll use the newer syntax, unless there's a reason we want to still run the tests in Go 1.16.

amchiclet: Oh, and some context where I read these things: [tracking bug](https://github.com/golang/go/issues/41184) and [transition plan](https://golang.org/s/go-build-design#transition).

amchiclet: Yep, things still work using just the newer syntax. Thanks for pointing it out.

PR #833, opened 2022-05-11
feat: library should be released as 4.2.0

amchiclet:
I think this previous [PR](https://github.com/googleapis/cloud-profiler-nodejs/pull/828) should have been a feature and the minor version should be bumped up as a result.

Creating an empty PR here with "feat:" in the title, so Release Please picks it up and the next release would be a minor version update.


amchiclet:
I think this [PR](https://github.com/googleapis/cloud-profiler-nodejs/pull/828) should have been a feature and the minor version should be bumped up as a result.

nolanmar511 (approved):

PR #835, opened 2022-05-16
build!: update library to use Node 12

sofisl:
Migrate library to Node 12


bcoe (approved):

nolanmar511 (approved):

alexander-fenster (approved):

alexander-fenster (approved):

system-test/busybench/package.json
```diff
@@ -20,5 +23,7 @@
     "posttest": "npm run check"
   },
   "devDependencies": {},
-  "dependencies": {}
+  "dependencies": {
+    "@google-cloud/profiler": "^4.2.0"
```
nolanmar511, line 27: I think this change means the system test will always run with @google-cloud/profiler 4.2.0?

If so, that would mean the system test won't work anymore

PR #851, opened 2022-08-15
DO_NOT_MERGE test: system test node 14 and 16

amchiclet:


nolanmar511 (approved):

nolanmar511:
LGTM; possible that Node 18 tests will fail b/c there aren't prebuilt binaries for Node 18.

amchiclet:
Yep, you're right. Pre-built is an issue for node 18. Scoping this PR to only include node 16.

nolanmar511 (approved):

weyert:
Any news on the status of this PR? Tomorrow Node v18 will become LTS and so far we haven't been able to use Profiling while I think it can be really   handy to have when trying to get to the bottom of performance issues on GKE

amchiclet:
There's [work](https://github.com/google/pprof-nodejs/pull/227) being done now that is needed to support Node 18 in pprof-nodejs (which this project depends on).

After that work is done, this project will then be updated to support up to Node 18. Unfortunately I don't have the capacity to continue this work and somebody else in the team will be picking it up.

weyert:
> There's [work](https://github.com/google/pprof-nodejs/pull/227) being done now that is needed to support Node 18 in pprof-nodejs (which this project depends on).
> 
> After that work is done, this project will then be updated to support up to Node 18. Unfortunately I don't have the capacity to continue this work and somebody else in the team will be picking it up.

Great, glad to see, work is in progress to get support for newer versions of Node :)

aabmass:
This was already addressed in https://github.com/googleapis/cloud-profiler-nodejs/pull/891

I will also add node 20 in that file now

PR #935, opened 2024-07-02
chore: upgrade kokoro test to use latest go version

aabmass:


aabmass:
Kokoro tests should pass now that https://github.com/googleapis/google-cloud-go/pull/10484 is merged. Just need Go to pick up the new HEAD ref

system-test/integration_test.sh
```diff
@@ -29,21 +29,19 @@ export GOOGLE_APPLICATION_CREDENTIALS="${SERVICE_KEY}"
...
 export PATH=$PATH:/usr/local/go/bin
+go version
 
 # Initializing go modules allows our dependencies to install versions of their
 # dependencies specified by their go.mod files. This reduces the likelihood of
 # dependencies breaking this test.
-go version
 go mod init e2e
-retry go get cloud.google.com/go/profiler/proftest@HEAD
+retry go get cloud.google.com/go/profiler/proftest@main
```
aabmass, line 44: For some reason `@HEAD` is pulling in the last release and not actually pulling from HEAD to pick up https://github.com/googleapis/google-cloud-go/pull/10484. This seems to fix it for now to actually pull in the latest.

psx95: Quick verification - Do we actually want to always fetch from latest commit instead of latest release?

psx95 (approved):
LGTM.

PR #402, opened 2019-02-12
chore: Remove hack for determining server-specified backoff.

nolanmar511:
Fixes #250 

Tested: Ran ~20 node jobs with matching service/version/labels. Confirmed that back off matched the back offs specified in the response body.

PR #960, opened 2025-04-09
security: replace vulnerable parse-duration with CJS compatible ms library

### LINKED ISSUE #953: Bump parse-duration to >=2.1.3
GitHub has the following security advisory for the `parse-duration` package that this library depends on

https://github.com/advisories/GHSA-hcrg-fc28-fcg5

This package currently depends on `parse-duration = ^1.0.0` but the patched version is in `^2.1.3`.

aabmass:
Fixes https://github.com/googleapis/cloud-profiler-nodejs/issues/953

Replace [parse-duration](https://www.npmjs.com/package/parse-duration) with [ms](https://www.npmjs.com/package/ms), which provides ESM and CJS builds.

BEGIN_COMMIT_OVERRIDE
fix: replace vulnerable parse-duration with CJS compatible ms library
END_COMMIT_OVERRIDE


src/profiler.ts
```diff
@@ -106,45 +106,20 @@ function getServerResponseBackoff(body: object): number | undefined {
...
+            return backoffMillis;
+          }
         }
       }
     }
   }
   return undefined;
 }
 
-/**
```
aabmass, line 124: This is dead code which was missed in https://github.com/googleapis/cloud-profiler-nodejs/pull/402

src/profiler.ts
```diff
@@ -111,7 +111,8 @@ function getServerResponseBackoff(body: object): number | undefined {
         item.retryDelay &&
         typeof item.retryDelay === 'string'
       ) {
-        const backoffMillis = parseDuration(item.retryDelay)!;
+        const retryDelay: `${number}s` = item.retryDelay;
```
aabmass, line 114: This should be safe because [JSON encoding of protobuf duration](https://protobuf.dev/programming-guides/json/#:~:text=Generated%20output%20always%20contains%200%2C%203%2C%206%2C%20or%209%20fractional%20digits%2C%20depending%20on%20required%20precision%2C%20followed%20by%20the%20suffix%20%22s%22.%20Accepted%20are%20any%20fractional%20digits%20(also%20none)%20as%20long%20as%20they%20fit%20into%20nano%2Dseconds%20precision%20and%20the%20suffix%20%22s%22%20is%20required.) is always suffixed with "s".

punya: Might be worth adding that info as a comment for future readers.

test/test-profiler.ts
```diff
@@ -919,48 +914,4 @@ describe('Profiler', () => {
       }
     );
   });
-  describe('parseBackoffDuration', () => {
```
punya, line 922: Why are these tests no longer necessary?

aabmass: I removed the tested function which was dead code missed in https://github.com/googleapis/cloud-profiler-nodejs/pull/402

src/profiler.ts
aabmass: Added

src/config.ts
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {GoogleAuthOptions} from '@google-cloud/common';
-import parseDuration from 'parse-duration';
+import * as ms from 'ms';
```
psx95, line 16: Nit: Why not use `import ms from 'ms'` ?

psx95 (approved):

aabmass:
I'm not sure why codecov is complaining since there are [tests covering this code](https://github.com/googleapis/cloud-profiler-nodejs/blob/ef0f58c0ac7abf15dcf3b2a31cd0a4b9bfdb3b72/test/test-profiler.ts#L835-L891) which break if I introduce a bug.

PR #964, opened 2025-04-09
test: add test for behavior with invalid duration returned

aabmass:
Follow up to https://github.com/googleapis/cloud-profiler-nodejs/pull/960. I verified this test behavior is the same before and after https://github.com/googleapis/cloud-profiler-nodejs/pull/960


psx95 (approved):
Might be good to indicate that expected/default backoff is 500ms (unless its already documented in source code).

